PR #22, opened 2024-06-10
[DeviceMemoryManager] Use offsets instead of pointers to locate tensors and fingerprints in device memory

jiazhihao:


wmdi (commented):
Can we have an individual offset manager for each mugraph, so that we can create multiple mugraphs at the same time?

wmdi (approved):

PR #23, opened 2024-06-11
[Graph Generator] Parallelize the generation algorithm

wmdi:
This PR parallelizes the search algorithm used in mugraph generation. Need to resolve conflicts with #22 .

Checkpoint is temporarily disabled.


include/mirage/search/search.h
```diff
@@ -89,46 +78,52 @@ class KernelGraphGenerator {
...
   char const *filename;
   std::vector<json> generated_graphs;
+  int num_thread;
+
+  std::chrono::milliseconds timeout;
 
 private:
   std::vector<std::shared_ptr<AlgebraicPattern>> final_patterns;
-  std::unordered_map<DTensor, std::shared_ptr<AlgebraicPattern>>
+  std::unordered_map<int, std::shared_ptr<AlgebraicPattern>>
```
jiazhihao: Is this still a map from dtensor to AlgebraicPattern? If so, shouldn't the key be int64_t?

include/mirage/search/search.h
```diff
@@ -89,46 +78,52 @@ class KernelGraphGenerator {
...
-                                 int depth);
   void optimize_layout(
       kernel::Graph &g, int op_idx, int ts_idx, int bop_idx, int bts_idx);
   void update_best_graph(kernel::Graph &g);
-  bool create_tb_outputs(SearchContext<STensor> &c,
-                         threadblock::Graph &g,
-                         int3 output_map);
+  bool create_tb_outputs(
+      SearchContext &c,
+      std::unordered_map<int, std::shared_ptr<AlgebraicPattern>> const
```
jiazhihao: Should the key be int or int64_t?

jiazhihao (approved):

include/mirage/search/search.h
wmdi: It should be int64_t. Pushed the updates.

include/mirage/search/search.h
wmdi: Updated

Issue #26, opened 2024-06-21
Support in-place optimization for threadblock elementwise binary operator


jiazhihao:
Threadblock elementwise binary operator currently allocates a separate shared memory tensor for the output, which results in high shared memory overhead. We should enable in-place optimization for elementwise binary operator (e.g., division) and close this issue once the implementation is merged to the main branch.

jiakunw:
I am not sure if I understood the code correctly, but I directly pushed the output tensot (which we used input1's mnemory space) to the output tensors

PR #29, opened 2024-07-06
Support in-place optimization for threadblock elementwise binary oper…

jiakunw:
…ator #26


jiakunw:
I am not sure if I understood the code correctly, but I directly pushed the output tensot (which we used input1's mnemory space) to the output tensors

PR #31, opened 2024-07-09
[KernelGraphGenerator] A few fixes

wmdi:
* Fix the segmentation fault and assertion failure caused by repeated guids
* Use multithreaded search by default
* Modify the timer to measure the wall time


src/kernel/customized.cc
jiazhihao, line 1: @wmdi Is it safe to comment out the following lines of code? Won't this cause memory leakage?

wmdi: Threadblock graph now has a destructor to free the operators

wmdi: Still, I think it is always better to avoid manually handling this via smart pointers. I will implement that in a separate pr.

jiazhihao: > Threadblock graph now has a destructor to free the operators

I see. That makes sense. I agree that it's better to avoid manual deletion of objects.

jiazhihao (approved):

PR #42, opened 2024-08-06
Improve parallel search performance

wmdi:
This PR improves the parallel search performance by

* Selecting a proper number of threads
* Shuffling the orders for search tree exploration for each thread for a better reuse of existing z3 invocations


wmdi:
Another modification included in this PR is solving memory leakage in search. I do not use smart pointers because the operators should be removed in a reversed order in the graph, which is not considered by the pointer reference manager. So currently I free the pointers in the destructor of graphs and disable the copy construction.

jiazhihao (approved):

PR #44, opened 2024-08-08
[Accumulator] This PR decouples accumulator from output saver in threadblock graphs

jiazhihao:
List of changes:
* A new `TBForloopAccumOp` is added to perform forloop accumulation, so the `TBOutputSaver` operator now only move result from smem to dmem without performing any accumulation.
* Each `STensor` now has a `after_accum` flag to indicate whether an accumulation has been applied. `TBOutputSaver` only accepts input `STensor` with `after_accum` set to true
* We now allocate a buffer in device memory to save stensors' fingerprints, which simplifies fingerprint calculation and allows Mirage to use higher precision integers for verification


wmdi (approved):

include/mirage/config.h
```diff
@@ -28,9 +28,13 @@ uint16_t const FP_PQ = 25651;
...
 // FP_Q_MUL_P_MOD_1 is a multiplier of Q and is 1 module P
 uint16_t const FP_Q_MUL_P_MOD_1 = 25425;
-size_t const MAX_SMEM_SIZE = 96 * 1024;                           // 96 KB
+size_t const MAX_SMEM_SIZE = 96 * 1024;                      // 96 KB
+// Note that we actually save stensors' fingerprints on GPU device memory
+// so MAX_SMEM_FP_SIZE can be larger than MAX_SMEM_SIZE
+size_t const MAX_SMEM_FP_SIZE = 1024 * 1024;                 // 1 MB
 size_t const MAX_DMEM_DATA_SIZE = (size_t)2 * 1024 * 1024 * 1024; // 2 GB
 size_t const MAX_DMEM_FP_SIZE = (size_t)2 * 1024 * 1024 * 1024;   // 2 GB
+size_t const MAX_NUM_THREADBLOCKS_PER_KERNEL = 1024;
```
interestingLSY, line 37: Probably 1024 threadblocks per kernel is too small to fully utilize the GPU...

interestingLSY: However, on the other side, if we set this to a larger number, the scratchpad memory for fingerprints calculation will be too big...

jiazhihao: Yeah, it's a trade-off between memory consumption and parallelism. This is also an overhead users have to pay when using mirage to search for mugraphs, so we probably don't want to make it very high. The current overhead is around 1GB for storing fingerprints.

PR #47, opened 2024-08-13
In place update for binary operators

jiakunw:
I removed uncessary code for memory allocation when returning result


src/threadblock/element_binary.cc
```diff
@@ -77,6 +77,13 @@ TBOperator *Graph::create_elementbinary_op(STensor const &input1,
   } else {
     return op;
   }
+<<<<<<< HEAD
```
jiazhihao: This is a git merge conflict you will need to fix. Make sure you can compile the code after the merge.

PR #48, opened 2024-08-13
Add more element_unary & element_binary operators at the kernel and threadblock levels

jiazhihao:


include/mirage/threadblock/cuda/element_unary.h
```diff
@@ -65,6 +70,19 @@ class TBElementUnaryFingerPrinter {
         result = (result * FP_Q_MUL_P_MOD_1) % FP_PQ;
         base_ptr[i] = result;
       }
+    } else if (type == mirage::type::TB_SILU_OP) {
+      for (int i = thread_id; i < num_elements; i += num_threads) {
+        FPType input = base_ptr[i];
+        // Note that we use $x * e^x$ as the fingerprint for SILU
+        // (i.e., $x / (1+e^{-x})$, since plus one can be easier
+        // implemented at any level of the GPU compute hierarchy
+        // and $x / e^{-x} = x * e^x$.
```
wmdi: Why do we need this workaround? Couldn't we just compute `x / (1+e^{-x})` directly?

include/mirage/type.h
```diff
@@ -48,6 +48,7 @@ enum KNOperatorType {
   KN_ADD_OP = 1008,
   KN_MUL_OP = 1009,
   KN_DIV_OP = 1010,
+  KN_SILU_OP = 1011,
   KN_ALLREDUCE_OP = 1011,
```
wmdi: `KN_SILU_OP` and `KN_ALLREDUCE_OP` are using the same id

include/mirage/threadblock/cuda/element_unary.h
jiazhihao: The main reason is to avoid introducing constant tensors (e.g., a tensor whose elements are all one or zero), but it should be fine to compute `x / (1+e^{-x})` as the fingerprint for silu.

jiazhihao: We will fix this approximation in a separate PR, which will use a more uniform approach to computing fingerprints.

include/mirage/config.h
```diff
@@ -20,18 +20,18 @@
 namespace mirage {
 namespace config {
 
-uint16_t const FP_P = 227;
-uint16_t const FP_Q = 113;
+uint16_t const FP_P = 167;
+uint16_t const FP_Q = 83;
```
wmdi, line 24: I think we can switch to a larger pair of P, Q given fingerprints are computed in device memory now. This can be a separate PR.

PR #50, opened 2024-08-17
[CUDA Transpiler] Supporting JIT transpilation and compilation

jiazhihao:


src/transpiler/resolve_tb_fusion.cc
```diff
@@ -0,0 +1,87 @@
...
+    for (tb::TBOperator *const op : tb_graph.operators) {
+      if ((op->op_type == type::TB_EXP_OP ||
+           op->op_type == type::TB_FORLOOP_ACCUM_NO_RED_OP)) {
+        tb::STensor const &input0 = op->input_tensors.at(0);
+        tb::TBOperator *prev_op = input0.owner_op;
+        // Don't fuse with an input op with forloop_dim = -1
+        if (prev_op->op_type == type::TB_INPUT_OP &&
+            dynamic_cast<tb::TBInputOp *>(prev_op)->forloop_dim == -1) {
+          continue;
+        }
```
xinhaoc, line 47: Do we still need this part in the case where there is already one in line 54

interestingLSY: Yes, this is indeed unnecessary, but I leave the code here as a reminder: we are never going to fuse with an input op with forloop_dim = -1

interestingLSY (approved):

xinhaoc (approved):
LGTM, Thanks for the excellent work!

PR #89, opened 2024-09-27
transpiler doc fix

xinhaoc:
figures and formulas at [here](https://mirage-project.readthedocs.io/en/latest/transpiler.html) are not rendered correctly
.I'm not sure how to make md file compatible with rst, maybe we should try creating a rst file for transpiler


jiazhihao (approved):

jiazhihao:
I like the idea of creating a new rst file for the transpiler. Thanks @xinhaoc

PR #113, opened 2024-10-22
Visualizer Ver1.0

NorthmanPKU:
**Description of changes:**
Embedded visualizer in python interface.
Simply use `graph.visualize("graph_name", "file_name")` to generate the graph.
Will support the newly added output node.

**Related Issues:**

Linked Issues:
- Issue #

Issues closed by this PR:
- Closes #


deps/cutlass
jiazhihao, line 1: We may want to update your cutlass commit. You can run `git submodule update` to update it.

NorthmanPKU: ok

NorthmanPKU: I have checkouted cutlass submodule to commit cc3c29a

python/mirage/_cython/core.pyx
```diff
@@ -101,6 +101,110 @@ bfloat16 = dtype('bf16')
 float32 = dtype('fp32')
 float64 = dtype('fp64')
 
+def get_kn_operator_type_string(int op_type):
+    if op_type == 1000:
```
jiazhihao: Instead of using absolute numbers, can we use `KN_XXX_OP` defined in `CCore.pxd`? This change will make it easier to update the encoding of KN/TB operators (e.g., we may need to change the ids corresponding to each operator type).

python/mirage/_cython/core.pyx
```diff
@@ -101,6 +101,110 @@ bfloat16 = dtype('bf16')
...
+    elif op_type == 1400:
+        return "kn_allreduce_op"
+    elif op_type == 1999:
+        return "kn_customized_op"
+    else:
+        return "unknown_op_type" + str(op_type)
+
+
+def get_tb_operator_type_string(int op_type):
+    if op_type == 2000:
```
jiazhihao: Similarly, let's also use `TB_XXX_OP` defined in `CCore.pxd`.

python/mirage/_cython/core.pyx
```diff
@@ -464,6 +833,32 @@ cdef class CyTBGraph:
         t = ctypes.cast(<unsigned long long>ptr, ctypes.c_void_p)
         return STensor(t)
 
+    property grid_dim:
```
jiazhihao, line 844: I was wondering if we can unify `grid_dim` and `block_dim` here. For example, both of them can return a dictionary or a tuple. Let's try not to use list.

python/mirage/kernel.py
```diff
@@ -333,3 +335,10 @@ def superoptimize(
                 best_graph, best_perf = g, perf
 
         return best_graph
+
+    def visualize(self, graph_name, file_name=None):
```
jiazhihao: I am not sure we need `graph_name` in this case.

jiazhihao (changes requested):
The PR looks great overall. I think it will be ready to be merged after addressing these minor issues. Thanks!

python/mirage/_cython/core.pyx
NorthmanPKU: It seems like block_dim is not used in the graph shown in the paper and actually also not used in the graph visualizer generated🤔. I removed it in current version.

jiazhihao (approved):

PR #139, opened 2024-11-15
[NKI] An initial implementation for the NKI Transpiler

jiazhihao:
**Description of changes:**

**Related Issues:**

Linked Issues:
- Issue #

Issues closed by this PR:
- Closes #

PR #146, opened 2024-11-25
[NKI transpiler] Fix forloop_accum_no_red_op acccumulator

praveen-velliengiri:
**Description of changes:**
This PR fixes the issue in transpiling `forloop_accum_no_red_op` by using correct accumulator name while transpiring.

Parent PR: https://github.com/mirage-project/mirage/pull/139

Also verified the functional correctness of generated NKI code with pytorch implementation,
Computed the absolute difference between the output of NKI code and vanilla pytorch RMSNorm implementation.
With a tolerance value = 0.001

Fraction of values within tolerance: 16.97%
Mean Absolute Error (MAE): 0.00000
Max Difference: 0.87500

NKI script file can be found here : [NKI rms_norm.py](https://gist.github.com/preejackie/c9c592a0dd7cde656a5176bbbd9ca07f)

### DIFF
--- a/src/nki_transpiler/transpile_tb.cc
+++ b/src/nki_transpiler/transpile_tb.cc
@@ -84,32 +84,36 @@ NKICustomOPTranspileResult
       // We determine whether we want to transpose the accum based on
       // tb_op's input. This is because, in the case tb_op's input
       // and output have different partition_dim, we defer the transpose
-      // to be performed after forloop to reduce cost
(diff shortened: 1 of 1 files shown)


jiazhihao:
@preejackie I had pushed similar fixes to `transpile_tb.cc`. Can you pull from the `nki` branch and resolve merge conflict?

PR #157, opened 2024-12-17
Triton

NorthmanPKU:
**Description of changes:**
1. MIRAGE_ROOT in `kernel.py` used to be `os.path.join(os.path.dirname(__file__), "include"`. Since `kernel.py` is under `mirage/python/mirage`, the correct form should be `os.path.join(os.path.dirname(__file__), "../../include"`.
3. add parameters for `superoptimizer()` in `kernel.py`
4. rm `tqdm` in loop and redundant notes 
5. add `tqdm` to `requirements.txt`
6. rm redundant output in `get_input_dim_expr()`
7. use single subprocess to run temperate codes for iterations

**Related Issues:**

Linked Issues:
- Issue #

Issues closed by this PR:
- Closes #


python/mirage/triton_profiler.py
```diff
@@ -114,39 +110,31 @@ def _run_profile(self, module_path: str, desc: str = "") -> float:
...
+                            )
+            
+            # Profile runs
+            starter = torch.cuda.Event(enable_timing=True)
+            ender = torch.cuda.Event(enable_timing=True)
+            
+            torch.cuda.synchronize()
+            starter.record()
+            
+            for _ in range(self.profile_iters):
```
jiazhihao: This implementation actually starting multiple subprocesses to run the kernel, and each subprocess run it once. Can we change to a new design that only launch a single subprocess in which the kernel is executed sequentially for multiple iterations.

jiazhihao (changes requested):
Other parts of the changes LGTM.

python/mirage/triton_profiler.py
NorthmanPKU: Done! Generated code will be like:
```
def run_kernel():
    # Warmup iterations
    for _ in range(2):
        device = torch.device('cuda')
        dtensor10133375 = torch.randn((16,4096,), dtype=torch.float16).to(device=device)
        dtensor10133376 = torch.randn((4096,6144,), dtype=torch.float16).to(device=device)
        dtensor10133377 = torch.randn((16,6144,), dtype=torch.float16).to(device=device)
        custom_kernel_1[(128, 1, 1)](dtensor10133377, dtensor10133375, dtensor10133376)
    
    # Profile iterations
    starter = torch.cuda.Event(enable_timing=True)
    ender = torch.cuda.Event(enable_timing=True)
    torch.cuda.synchronize()
    starter.record()
    for _ in range(5):
        device = torch.device('cuda')
        dtensor10133375 = torch.randn((16,4096,), dtype=torch.float16).to(device=device)
        dtensor10133376 = torch.randn((4096,6144,), dtype=torch.float16).to(device=device)
        dtensor10133377 = torch.randn((16,6144,), dtype=torch.float16).to(device=device)
        custom_kernel_1[(128, 1, 1)](dtensor10133377, dtensor10133375, dtensor10133376)
    ender.record()
    torch.cuda.synchronize()
    elapsed_time = starter.elapsed_time(ender) / 5
    print(f'Average time per iteration: {elapsed_time:.3f} ms')

if __name__ == "__main__":
    print("Running kernel...")
    run_kernel()
```

jiazhihao: Can users can `custom_kernel_1` as a function in the main Python program? Ideally, we want `superoptimize` to return a mugraph that can be directly executed by the caller (see `demo/rms_norm.py` as an example).

NorthmanPKU: Yes, since it returns the same class instance just as the cuda backend superoptimizer part

NorthmanPKU: Oh not yet

NorthmanPKU: 1. Support directly calling graph as funcs 
1.1. add execute_mugraph()in generated code 1.2. change __call__ to support several backends 1.3. other relative modify 
2. fix some errors in triton transpiler including mask form and matmul op

jiazhihao (approved):

PR #181, opened 2025-02-10
Triton warnings fix & triton kernel added

NorthmanPKU:
**Description of changes:**
1. Remove triton.ops dependency (except for ops.matmul)
2. Add triton kernel file as runtime file and logic to read kernels from it
3. Handle warnings for type mismatch and switch (some new ops are listed as notes to be further filled when merged with new_ops branch)
4. Fix exeucte_mugraph enter function logic related to input, middle and output tensors

**Related Issues:**

Linked Issues:
- Issue #

Issues closed by this PR:
- Closes #


benchmark/gated_mlp.py
jiazhihao, line 1: Let's add an input argument to control which backend to use.

benchmark/lora.py
jiazhihao, line 1: Same as above

jiazhihao (changes requested):

benchmark/gated_mlp.py
NorthmanPKU: 🆗

benchmark/lora.py
NorthmanPKU: Added more parsers to all these benchmarks

jiazhihao (approved):

PR #218, opened 2025-03-24
Triton kernel for RoPE

NorthmanPKU:
**Description of changes:**
1. Add rope.py including rope triton kernel
2. Leave Timing util class for now for possible future profiling
3. Triton will suffer from catastrophic warmup problem, we may need to add warmup at some suitable places

This snapshot shows time consumed in each calling of the kernel, where Duration 0 takes way more time, burdening the whole average result:
455661742801190_ pic

4. The output of the demo is only "!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!" (even before adding the new kernel). The reason should be further explored.
Screenshot 2025-03-24 at 04 34 09

**Related Issues:**

Linked Issues:
- Issue #

Issues closed by this PR:
- Closes #


jiazhihao:
The output should be normal text instead of repeated exclamation marks. What's the CUDA version you used?

NorthmanPKU:
Although reason for wrong graphs in checkpoints is still under examine, experiments show it's not a problem with this triton kernel since they can output exactly the same correctly with functional checkpoint graph candidates.
Screenshot 2025-03-27 at 16 07 03
Screenshot 2025-03-27 at 16 11 12

jiazhihao (approved):

PR #219, opened 2025-03-24
Adding operators with more than one output tensors: chunk op

AMKCode:
**Description of changes:**
Added the chunk operator that is able to split tensors into 2 equally sized outputs along a specified axis. Splitting into more than 2 outputs is currently not supported (chunk_size can only be 2).

Modified the Llama 8B example to demonstrate the use of the chunk operator.

**Related Issues:**

Linked Issues:
- Issue #

Issues closed by this PR:
- Closes #


AMKCode:
Chunk op is currently working at the kernel and threadblock level. However, search needs to be updated in order to support its use in search.

AMKCode:
Chunk operator is ready. CICD run fails but it appears to be an issue with the environment and not with code.

Issue #262, opened 2025-05-06
Support Mirage on Blackwell GPU Architecture


yzhou442:
Currently Mirage only supports Hopper and previous GPU architecture. Promoting Mirage to support Blackwell will enable more research opportunities on FP4 quantization.

PR #263, opened 2025-05-06
Blackwell Support

yzhou442:
**Description of changes:**
Promote Mirage to Blackwell GPU Architecture, some major new features introduced in Blackwell are:

1. **New UMMA instructions**: Core MMA instructions are updated with UMMA, which can support low precision data types including FP4 and FP6.
2. **New On-chip Tensor Memory for UMMA**: Used for MMA accumulation by replacing the role of registers.

**Related Issues:**

Linked Issues:
- Issue [#262 ](https://github.com/mirage-project/mirage/issues/262)

Issues closed by this PR:
- Closes #


src/transpiler/transpiler_kn.cc
```diff
@@ -517,102 +522,302 @@ TranspileResult Transpiler::transpile_ugraph() {
...
               tma_tmps.append(", ");
               m_inputs.append(", ");
+              tma_dst_pipe_layouts.append(", ");
+            }
+            if (tmaParams.m_input) {
+              tiled_mmas.push_back(tmaParams.tiled_mma);
             }
           }
+          // init tiled_mmas
+          if (config.target_cc == GPU_CC::B200 && use_2sm_mma) {
```
xinhaoc, line 734: Do you think we can refactor this part into separate functions for different archs?

src/transpiler/plan_stensor_memory.cc
```diff
@@ -255,11 +255,17 @@ AllocResult plan_memory(vector<TensorDecl> const &tensor_decls) {
 
 TBMemoryPlan Transpiler::get_threadblock_memory_plan(tb::Graph const &tb_graph,
                                                      TBSched const &tb_sched,
-                                                     bool hopper_arch) {
+                                                     bool hopper_arch,
```
xinhaoc, line 258: same with this function

src/transpiler/transpiler_kn.cc
yzhou442: Yes, it can be clearer if we create different functions and just call the function here

src/transpiler/plan_stensor_memory.cc
yzhou442: For this memory plan function, blackwell only has to plan several more mbarriers inside the shared memory compared to other archs, so the main body is actually the same, I don't think it is quite necessary to separate this function.

jiazhihao:
@yzhou442 Since most of your changes are to the transpiler, can you update this PR to be merged to the `mpk` instead of `main` branch? This can help minimize divergence between main and mpk and we will merge them into one in the near future.

yzhou442:
Sure

PR #272, opened 2025-05-19
multi patterns as inputs for check_pattern() and egg library for subpattern checking

Xiaoyu1125:
**Description of changes:**
In search.cc part, I changed the input pattern from shared_ptr<AbstractExpr> to std::vector<shared_ptr<AbstractExpr>>. In egg library, it will create one graph(which costs long time) and we can check all the subexpr in this graph at once. So I changed the input in the multi way.

In the subexpr checking, I use egg library in RUST. In this part, the main idea is to form the e-graph of final pattern and use lookup_expr() to check whether the subexpr is the subgraph of this e-graph.

The changes in setup.py and CMakeLists.txt aim to adapt the  RUST func to cpp and python version.

By the way, I add a boolean parameter to switch the verification type, probabilistic and formal verification.

Performance: Now, we can support all the integers( rather than only power of 2), and there is a strong speedup in the search part. These two figures show the comparison in the demo_group_query_attention.py. The first graph in the new version. The second is the origin mirage.
屏幕截图 2025-05-19 155608
屏幕截图 2025-05-19 164340

**Related Issues:**

Linked Issues:
- Issue #

Issues closed by this PR:
- Closes #


wmdi:
Can you run format?

demo/verification_tests/gated_mlp_verifier.py
wmdi, line 1: I don't quite understand the logic of this script... It seems that none of `torch_gated_mlp`, `graph`, or `xx_output` is used. Also, I don't think the correctness of formal verifier can be tested by just comparing the optimal mugraph discovered. It is neither sufficient or necessary (the variance of profiling may change the optimal one).

I know that you have actually compared the results of prob verifier and formal verifier for every mugraph generated, which is the true test. But every script in the demo dir should demonstrate the correct usage of some functionalities. I would suggest just remove or redesign these scripts.

deps/Rust/src/abstract_subexpr-stamp/abstract_subexpr-build-.cmake
wmdi, line 1: `deps` should only contain third-party dependencies, usually as submodules. Please reorganize the file structure.

deps/Rust/src/abstract_subexpr-stamp/abstract_subexpr-build-err.log
wmdi, line 1: The log files should be removed?

include/mirage/search/abstract_expr/abstract_expr.h
```diff
@@ -20,9 +31,9 @@ class AbstractExpr {
   virtual z3::expr
       to_z3(z3::context &c,
             std::unordered_set<std::string> &all_variables) const = 0;
-  bool subpattern_to(AbstractExpr const &other) const;
-  bool operator==(AbstractExpr const &other) const;
+  std::unordered_map<int, bool> subpattern_to(const std::vector<std::shared_ptr<AbstractExpr>>& input_patterns) const;
```
wmdi: I think using `vector<bool>` as the return value makes more sense.

include/mirage/search/search.h
```diff
@@ -72,7 +76,7 @@ class KernelGraphGenerator {
       size_t depth);
 
   void preprocess(kernel::Graph const &computation_graph);
-  bool check_pattern(std::shared_ptr<AbstractExpr> pattern);
+  std::unordered_map<int, bool> check_pattern(std::vector<std::shared_ptr<AbstractExpr>>& inputs);
```
wmdi: use vector

include/mirage/triton_transpiler/runtime/triton_kernels.py
wmdi, line 1: The diff shows this file is deleted. Did you merge main into your branch?

src/search/abstract_expr/abstract_subexpr/Cargo.lock
wmdi, line 1: I don't think this file should be included?

src/search/search.cc
```diff
@@ -145,28 +147,39 @@ void KernelGraphGenerator::generate_next_operator(
...
-
-          if (new_op) {
-            c.kn_graph->operators.push_back(new_op);
-            if (check_range(init_ranges, target_ranges, *c.kn_graph)) {
-              if (depth < max_depth) {
-                num_tasks++;
-                SearchContext c_tmp = SerializedSearchContext(c).deserialize();
+        }
+        std::unordered_map<int, bool> results = check_pattern(inputs);
+        // elapsed += std::chrono::steady_clock::now() - begin_time;
```
wmdi: Please remove all the commented codes

wmdi (changes requested):

include/mirage/search/abstract_expr/abstract_expr.h
Xiaoyu1125: I should use the index(the first value of the map) to find the corresponding input_tensors

demo/verification_tests/gated_mlp_verifier.py
Xiaoyu1125: I have deleted these demo, since they are no longer necessary.

deps/Rust/src/abstract_subexpr-stamp/abstract_subexpr-build-.cmake
Xiaoyu1125: removed

deps/Rust/src/abstract_subexpr-stamp/abstract_subexpr-build-err.log
Xiaoyu1125: removed

src/search/abstract_expr/abstract_subexpr/Cargo.lock
Xiaoyu1125: removed

src/search/search.cc
Xiaoyu1125: done

Xiaoyu1125:
image
done

include/mirage/search/abstract_expr/abstract_expr.h
wmdi: You will not lose any information when you use vector, since a vector, for example, [result1, result2, result3] is equivalent to a map {{0, result1}, {1, result2}, {2, result3}}. This modification only needs a few lines of changes in `search.cc`.

Xiaoyu1125: done

include/mirage/triton_transpiler/runtime/triton_kernels.py
Xiaoyu1125: I have restored this file

include/mirage/search/abstract_expr/abstract_expr.h
wmdi: And for `AbstractExpr::subpattern_to`

Xiaoyu1125: done

PR #293, opened 2025-06-07
support more `REDUCTION_SIZE`

undefined-c0der:


CMakeLists.txt
jiazhihao, line 1: Let's not set Z3 paths here since the pip install path sets Z3 paths in a different way

cpp_examples/runtime_demo/runtime_demo.cu
jiazhihao, line 1: This file was deprecated.

jiazhihao (changes requested):
See my comments inlined.

PR #296, opened 2025-06-10
align `norm_linear` task in Mirage Runtime framework

undefined-c0der:


include/mirage/persistent_kernel/persistent_kernel.cuh
```diff
@@ -390,13 +393,85 @@ __global__ void persistent_kernel(RuntimeConfig config) {
...
+                  task_desc.inputs[1].base_ptr,
+                  task_desc.inputs[2].base_ptr,
+                  1e-6f, // eps
+                  task_desc.outputs[0].base_ptr);
+            } else {
+              printf("Unsupported RMSNorm linear task stride: %d\n",
+                     task_desc.outputs[0].stride[0]);
+              assert(false && "Unsupported RMSNorm linear task");
+            }
+            // if (threadIdx.x == 0) {
```
jiazhihao: Can we remove these comments before merging?

tests/runtime_python/runtime_kernel_wrapper.cu
```diff
@@ -228,7 +228,11 @@ __global__ void norm_linear_kernel_wrapper(void const *input_ptr,
                                            void const *weight_ptr,
                                            float eps,
                                            void *output_ptr) {
-  norm_linear_task_impl<T, BATCH_SIZE, OUTPUT_SIZE, REDUCTION_SIZE>(
+  norm_linear_task_impl<T,
```
jiazhihao, line 231: Why we have two `OUTPUT_SIZE`s in the template argument?

undefined-c0der: The second one is treated as `O_STRIDE` here

include/mirage/persistent_kernel/tasks/embedding.cuh
xinhaoc, line 1: Does this include negative numbers? If not, can we use usigned 32 int data type

jiazhihao: Torch used int64_t for token IDs, so I suggest that we use the same data type as torch.

jiazhihao (approved):

PR #317, opened 2025-06-19
[MPK] Persistent Kernels in Mirage

jiazhihao:
**Description of changes:**

Merging persistent kernel into the main branch of Mirage.

**Related Issues:**

Linked Issues:
- Issue #

Issues closed by this PR:
- Closes #


src/threadblock/graph.cc
xinhaoc, line 1: Is this the only place where "store_in_dmem" works? Not quite understand about this part

cmake/cuda.cmake
xinhaoc, line 1: https://cmake.org/cmake/help/latest/module/FindCUDA.html#module:FindCUDA
do you think if we can update it to the newest version

src/threadblock/graph.cc
jiazhihao: Only MPK involves cases where `store_in_dmem` is set to be true. This is because for embedding, we need to load the entire vocabulary to compute the output but the vocabulary does not fit in smem.

cmake/cuda.cmake
jiazhihao: Yes, it's a great idea to update it.

wmdi (approved):

Issue #336, opened 2025-06-22
Support Qwen3 demo with other Qwen3 model and other GPUs


ywq880611:
Hi MPK Team,

Awesome work on the MPK project – it's very interesting!

I understand the current demo maybe just support **Qwen3-8B** on **A100**, right? I've been trying to get the `python demo/qwen3/demo.py` running. Given my **RTX 3080** (10GB VRAM), I anticipated challenges with larger models like **8B**, I tried some smaller **Qwen3** model, but I also noticed some errors and what appears to be hardcoded shape and size in both the **python and cpp/CUDA** parts that are causing the demo to fail on some small model, like **Qwen3-0.6B**.

I'd like to help improve this. I plan to prepare and submit some PRs to fix these issues and hopefully expand the demo's compatibility with more models and GPUs.

Thanks again for such an fantastic project!

jiazhihao:
@ywq880611 Thanks for the PR and for your interest in our project. You are right that the current demo is hardcoded for Qwen3-8B. Your PRs are very helpful to make the demo more general and supports other models. If you are interested in contributing to the project, make sure to join our slack channel where we discuss our development roadmap and ongoing tasks.

RoadTLife:
Is it possible to use multiple GPUs for the inference of the 32b model? Could you provide the list of currently supported models and the method for performing inference using multiple GPUs?

izhuhaoran:
After #544 , all qwen3 models (Qwen3-0.6B, Qwen3-1.7B, Qwen3-4B, Qwen3-8B, Qwen3-14B, Qwen3-32B) test pass on demo_hopper.py

PR #338, opened 2025-06-22
Fix some hardcode

ywq880611:
**Description of changes:**
There are some hardcode for **Qwen3-8B** and **A100**, this PR supports more models and GPUs, but there're something still in draft stage, like to how define `num_workers`, `num_local_schedulers` and `num_remote_schedulers` automatically, any insights would be very appreciated! 

**Related Issues:**

Linked Issues:
- Issue #336

### DIFF
--- a/demo/qwen3/demo.py
+++ b/demo/qwen3/demo.py
@@ -6,6 +6,13 @@ import torch.distributed as dist
 import argparse
 import os
 
+def grid_for_rmsnorm_linear_layer(size):
+    # 96 and 64 are enough to cover all Qwen3 model?
+    if size % 96 == 0:
+        return 96
+    elif size % 64 == 0:
+        return 64
+
 if __name__ == "__main__":
     parser = argparse.ArgumentParser()
     parser.add_argument("--use-mirage", action="store_true", help="Use Mirage kernels")
@@ -35,7 +42,8 @@ if __name__ == "__main__":
     print("Input arguments:", args)
     print(f"world_size({world_size}) rank({rank})")
     # model_name = "Qwen/Qwen2.5-7B-Instruct"
-    model_name = "Qwen/Qwen3-8B"
+    # TODO(Wenqin): set it as 8B befor landing.
+    model_name = "Qwen/Qwen3-0.6B"
     torch.set_default_dtype(torch.bfloat16)
 
     torch.cuda.set_device(rank)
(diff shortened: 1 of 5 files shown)


demo/qwen3/demo.py
```diff
@@ -31,7 +38,8 @@
     print("Input arguments:", args)
     print(f"world_size({world_size}) rank({rank})")
     # model_name = "Qwen/Qwen2.5-7B-Instruct"
-    model_name = "Qwen/Qwen3-8B"
+    # TODO(Wenqin): set it as 8B befor landing.
```
jiazhihao: Let's make 8B the default modal and allow users to pass a `--model` argument.

demo/qwen3/demo.py
```diff
@@ -107,12 +115,19 @@
             ).contiguous()
         else:
             profiler_tensor = None
+        # TODO(Wenqin): How to set num_local_schedulers and num_remote_schedulers? 
+        # Maybe we should remove the below change and let user set it manully?
+        num_workers = 64
```
jiazhihao: Instead of hardcode number of workers/schedulers, we can add default configurations in a separate python file that map each gpu type to a worker/scheduler configuration. This should be done in a separate PR.

wenqiny: Thanks for the suggestion, I could draft another PR to do it later, but there is a question from my side:

Is there any hint or heuristic the worker configuration? IIUC, if the `rank` is `0`, the `num_remote_schedulers` should be `0` too, right? but if `rank` is not `0`, how we set `num_local_schedulers` and `num_remote_schedulers`? Or we just support `rank` equals to `0` yet? Any insights would be very appreciated!

ywq880611:
Addressed comments, PTAL, thanks!

demo/qwen3/demo.py
jiazhihao: We will unify `num_local_schedulers` and `num_remote_schedulers` as `num_schedulers` to simplify the API. In this case, the total number of GPUs will not affect the configuration.

demo/qwen3/demo.py
wenqiny: Done

demo/qwen3/demo.py
wenqiny: Got it, thank you for the explanation!

ywq880611 (commented):
Addressed comments, is there anything left we should do in this PR?

jiazhihao:
@ywq880611 Can you merge updates to mpk to your branch? This will help pass CI.

PR #380, opened 2025-07-03
feat: Add sine and cosine operations to Mirage framework

EpicUsaMan:
**Description of changes:**
Sin & Cos activation

- Many models with image modality use it
- FANFormer use it https://arxiv.org/abs/2502.21309

**Related Issues:**

Linked Issues:
- Issue #

Issues closed by this PR:
- Closes #


EpicUsaMan:
Looks like I forgot something which was not caught by local build

Will fix it and write proper tests

include/mirage/utils/fingerprint_functions.h
```diff
@@ -96,6 +96,28 @@ inline __device__ FPType compute_gelu_fingerprint(FPType input,
   return result;
 }
 
+inline __device__ FPType compute_sin_fingerprint(FPType input) {
+  // Simple approximation: use the input modulo some value to simulate sin behavior
+  uint32_t q_residual = input % FP_Q;
+  uint32_t p_residual = input % FP_P;
+  // Simulate sin by using some transformation of the input
+  q_residual = (q_residual * 7) % FP_Q;  // Simple transformation
```
wmdi, line 104: I think this simulation is too simple and can be easily confused with scalar multiplication, which may cause false positive. Could you change to a better one?

eitanturok:
Hi! Big fan of mirage here. I have a quick question.

In the [mirage paper](https://arxiv.org/pdf/2405.05751), the probabilistic equivalence verifier only guarantes the kernel is correct with high probability on LAX mu-graphs. Do cos and sin fall under the definition of a LAX mu-graph? (Recall defintion 5.1 from the paper states the LAX mu-graph contains only multi-linear operators, division, and at most exponentiation.)

Thanks.

PR #392, opened 2025-07-09
Blackwell Transpiler Support

yzhou442:
**Description of changes:**
Update the transpiler for Blackwell-specific code generation.

**Related Issues:**

Linked Issues:
- Issue #

Issues closed by this PR:
- Closes #


src/transpiler/transpiler_tb_blackwell.cc
```diff
@@ -0,0 +1,1576 @@
...
+
+  bool need_mbarrier = false;
+  string tmem_base_ptr_name = "tmem_base_ptr";
+  string mbarrier_ptr_name = "mma_barrier_ptr";
+
+  size_t barrier_addr = mem_plan.smem_size;
+  for (auto [guid, addr] : mem_plan.addrs) {
+    if (guid == mem_plan.tmem_base_ptr_guid) {
+      code.e("uint32_t *$ = (uint32_t*)(buf + $);", tmem_base_ptr_name, addr);
+    } else if (guid >= mem_plan.mbarrier_buf_guid_offset) {
```
xinhaoc, line 431: do we need to store this mbarrier_buf_guid_offset as a part of mem_plan? it looks like a const value

yzhou442: Yes, it is a const value that is used to identify the mbarrier ptr from the other stensors. Where do you suggest to store it?

xinhaoc: I just realize you're using this as an offset of the barrier pointers, it's good to keep it unchanged

jiazhihao:
It seems the build currently fails with the following error message 
```

xinhaoc (approved):

PR #414, opened 2025-07-19
Fix CPU finger_prints python build

praveen-velliengiri:
This patch fixes the python `pip install` build issue of CPU only finger prints related work.


praveen-velliengiri:
> @praveen-velliengiri Can you merge this PR to `mpk` branch instead of `search`? All search changes have been integrated into mpk.

@jiazhihao Actually mpk doesn't have changes related to cpu fingerprints. @wmdi committed those changes to search branch, it was planned that after the python build work got completed we merge to mpk.

PR #422, opened 2025-07-21
Implementing float4 struct to init zero_buffer

Hengrui-Chen:
**Description of changes:**

Implemented float4 struct to init zero_buffer

**Related Issues:**

Linked Issues:
- Issue #281 

Issues closed by this PR:
- Closes #

### DIFF
--- a/demo/qwen3/demo.py
+++ b/demo/qwen3/demo.py
@@ -330,7 +330,7 @@ if __name__ == "__main__":
             input=x, 
             weight=w, 
             output=y, 
-            grid_dim=(max_factor_leq_n(4096, 96 // total_tokens_per_iter), total_tokens_per_iter, 1), 
+            grid_dim=(max_factor_leq_n(hidden_size, 96 // total_tokens_per_iter), total_tokens_per_iter, 1), 
             # grid_dim=(1, 1, 1), 
             block_dim=(128, 1, 1),
             input_source=(spec_decode_config is not None) # 0: all_tokens, 1: input_token (spec decoding)
--- a/include/mirage/persistent_kernel/persistent_kernel.cuh
+++ b/include/mirage/persistent_kernel/persistent_kernel.cuh
@@ -185,9 +185,40 @@ __device__ void terminate_schedulers(RuntimeConfig config) {
   }
 }
 
-__global__ void persistent_kernel(RuntimeConfig config) {
-  __shared__ TaskId cur_task_id;
-  __shared__ TaskDesc task_desc;
+__device__ void worker_checker(RuntimeConfig config) {
+  assert(gridDim.y == 1);
+  assert(gridDim.z == 1);
+  // Each worker SM serves a single worker
+  // Each scheduelr SM serves four schedulers
+  int num_schedulers =
+      config.num_local_schedulers + config.num_remote_schedulers;
+  assert(num_schedulers % 4 == 0);
+  assert(gridDim.x == config.num_workers);
+  assert(config.num_workers <= MAX_NUM_WORKERS);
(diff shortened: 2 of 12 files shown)


jiazhihao (approved):

Hengrui-Chen:
> @Hengrui-Chen Why the implementation uses float4 instead of uint128?

In the implementation, I initially chose `float4` over `uint128` for zero initialization because the original context (Issue #281) suggested to use `float4`. On the other hand, I recalled that `float4` is a built-in CUDA vector type, well designed for aligned, coalesced memory operations, so I thought it would be ok. I have written and run some code to compare the performance of initialization using float4 and uint128. And the test result shows that the latter is slightly faster. (for 2^20 initializations, float4 zero init took 0.037 ms, __uint128_t zero init took 0.028 ms). Given the negligible performance difference, I think we can adapt to whichever approach. But since using uint128 aligns better with the project’s conventions and is preferred elsewhere, I’m happy to update the implementation accordingly. Below is the performance test code for your reference.
```cuda
#include <cuda_runtime.h>
#include <cstdio>

#define CHECK_CUDA(call) \
    do { \
        cudaError_t err = call; \
        if (err != cudaSuccess) { \
            fprintf(stderr, "CUDA Error at %s:%d - %s\n", __FILE__, __LINE__, cudaGetErrorString(err)); \
            exit(EXIT_FAILURE); \
        } \
    } while (0)

constexpr size_t NUM_ELEMENTS = 1 << 20; // 1M elements
constexpr size_t BUFFER_SIZE = NUM_ELEMENTS * sizeof(float); // Total bytes
__device__ float4 zero;
float4 zero_ = make_float4(0.f, 0.f, 0.f, 0.f);

// Method 1: float4 initialization
__global__ void init_with_float4(float* buffer) {
    size_t idx = threadIdx.x + blockIdx.x * blockDim.x;
    if (idx * 4 >= NUM_ELEMENTS) return;

    ((float4*)buffer)[idx] = zero;
}

// Method 2: __uint128_t initialization (only works if sizeof(float)*4 == 16)
__global__ void init_with_uint128(float* buffer) {
    size_t idx = threadIdx.x + blockIdx.x * blockDim.x;
    if (idx * 4 >= NUM_ELEMENTS) return;
    ((__uint128_t*)buffer)[idx] = 0ul;
}

void benchmark(void (*kernel)(float*), const char* name) {
    float* d_buffer;
    CHECK_CUDA(cudaMalloc(&d_buffer, BUFFER_SIZE));
    CHECK_CUDA(cudaMemcpyToSymbol(zero, &zero_, sizeof(float4), 0, cudaMemcpyHostToDevice));


    cudaEvent_t start, stop;
    CHECK_CUDA(cudaEventCreate(&start));
    CHECK_CUDA(cudaEventCreate(&stop));

    // Launch configuration
    dim3 block(256);
    dim3 grid((NUM_ELEMENTS / 4 + block.x - 1) / block.x);

    // Run kernel and time it
    CHECK_CUDA(cudaEventRecord(start));
    kernel<<<grid, block>>>(d_buffer);
    CHECK_CUDA(cudaEventRecord(stop));
    CHECK_CUDA(cudaEventSynchronize(stop));

    float ms = 0;
    CHECK_CUDA(cudaEventElapsedTime(&ms, start, stop));

    printf("%s took %.3f ms\n", name, ms);

    cudaFree(d_buffer);
    cudaEventDestroy(start);
    cudaEventDestroy(stop);
}

int main() {
    printf("Benchmarking initialization methods...\n");

    benchmark(init_with_float4, "float4 zero init");
    benchmark(init_with_uint128, "__uint128_t zero init");

    return 0;
}
```

Issue #520, opened 2025-10-08
[Bug] - Attention Kernel Randomness


xinhaoc:
### Bug Description

The result of our attention kernel is not 100% same with the same prompt and greedy sampling in multi-iterations.

### Build Environment

run the demo.py

### Error Message Output

output

PR #536, opened 2025-10-16
Optimize multitoken_paged_attention_hopper.cuh && linear_swapAB_hopper.cuh

yzhou442:
**Description of changes:**

- Replace TMA_K/V with `load_smem` like flashinfer.
- Keep TMA_Q as an optional choice, since TMA_Q and `load_smem` give similar performance in both demo and unit est profiling.
- Reduce share memory usage by reusing space.
- Eliminate randomness by fixing sync issues in `multitoken_paged_attention_hopper.cuh`, now hopper demo always generates identical results. #520  
- Use smaller mma atom size (64x8x16) in `linear_swapAB_hopper.cuh` to handle batch size <= 8.

**Related Issues:**

Linked Issues:
- Issue #

Issues closed by this PR:
- Closes #

### DIFF
--- a/include/mirage/persistent_kernel/tasks/hopper/multitoken_paged_attention_hopper.cuh
+++ b/include/mirage/persistent_kernel/tasks/hopper/multitoken_paged_attention_hopper.cuh
@@ -750,14 +750,17 @@ __device__ __forceinline__ void multitoken_paged_attention_hopper_impl(
 
     // store the output
 #if OUTPUT_PTR
+#pragma unroll
     for (int elem_idx = threadIdx.x;
-         elem_idx < num_tokens * NUM_QO_PER_KV * HEAD_DIM;
+         elem_idx < num_tokens * NUM_QO_PER_KV * HEAD_DIM / CP_CHUNK_SIZE;
          elem_idx += NUM_THREADS) {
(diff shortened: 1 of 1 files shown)


jiazhihao (approved):

include/mirage/persistent_kernel/runtime_header.h
jiazhihao, line 1: I think we should assume that MPK_TARGET_CC is always set and assert failure if it wasn't defined in this header.

yzhou442: You are right, this change was intended for making unit test to correctly define `MAX_DYNAMIC_SHARED_MEMORY_SIZE`, I have changed to pass the `-DMPK_TARGET_CC=90` inside the unit test

include/mirage/persistent_kernel/tma.cuh
jiazhihao, line 1: Why we changed TILE_SIZE from 256 to 128?

yzhou442: Using smaller TILE_SIZE in `linear_swapAB_hopper` can save some share memory usage, and also I found that 128 is a good balance point between computation and data loading, which gives us better performance while less share memory usage

PR #544, opened 2025-10-21
feat: support more hidden_dim in rms_norm to enable all Qwen3 models for demo_hopper

### LINKED ISSUE #538: An assertion issue in rmsnorm_hopper.cuh
### Bug Description

1. I've discovered that when the model's hidden size isn't a multiple of 2048, the compilation process triggers an assertion in rmsnorm. Currently, I can only attempt to modify the tiling method to match my model's hidden size. Is there a better way to adapt this? 
rmsnorm_hopper.cuh
`constexpr int CHUNK_SIZE = 16 / sizeof(T); // 128b copy-async
  constexpr int TILE_SIZE = NUM_THREADS * CHUNK_SIZE;
  static_assert(HIDDEN_DIM % TILE_SIZE == 0);`
My model's hidden size is 2560.

### Build Environment

run our_demo.py

### Error Message Output

assertion issue

izhuhaoran:
**Description of changes:**

The current code can only run successfully with Qwen3-8B and Qwen3-1.7B for demo_hopper.py. However, for Qwen3-0.6B, Qwen3-4B, and Qwen3-32B, it fails due to the `static_assert(HIDDEN_DIM % TILE_SIZE == 0)` check in rms_norm. This PR references the cute implementation:

https://github.com/NVIDIA/cutlass/blob/main/include/cute/arch/copy_sm80.hpp#L55-L65

It modifies the `load_smem` implementation by using `cp.async.ca.shared.global.L2::128B`,  which supporting more cp-sizes, thereby enabling Qwen3-0.6B, Qwen3-4B, and Qwen3-32B. Additionally, other common models' hidden_dim should also be supported as a result of this change.

**Related Issues:**

Linked Issues:
- Issue #

Issues closed by this PR:
- Closes #538

### DIFF
--- a/include/mirage/persistent_kernel/tasks/common/copy_sm80.cuh
+++ b/include/mirage/persistent_kernel/tasks/common/copy_sm80.cuh
@@ -38,17 +38,20 @@ __device__ __forceinline__ void cp_async_fence() {
 #endif
 }
 
-// load 128 bytes values from global to shared memory async
-template <typename T>
+// load BYTES (and prefetch 128 bytes) from global to shared memory async
+template <typename T, int BYTES = 16>
 __device__ __forceinline__ void load_smem(T *smem_ptr, T const *gmem_ptr) {
 #ifdef CP_ASYNC_SM80_ENABLED
+  static_assert(BYTES == 4 || BYTES == 8 || BYTES == 16,
+                "cp.async only supports 4, 8, or 16 bytes");
+
   uint32_t smem_int_ptr =
       static_cast<uint32_t>(__cvta_generic_to_shared(smem_ptr));
-  asm volatile("cp.async.cg.shared.global.L2::128B [%0], [%1], %2, %3;\n" ::"r"(
+  asm volatile("cp.async.ca.shared.global.L2::128B [%0], [%1], %2, %3;\n" ::"r"(
                    smem_int_ptr),
                "l"(gmem_ptr),
-               "n"(16),
-               "r"(16));
+               "n"(BYTES),
+               "r"(BYTES));
 #endif
 }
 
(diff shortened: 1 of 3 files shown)


izhuhaoran:
@jiazhihao would you please review this pr?

include/mirage/persistent_kernel/tasks/common/copy_sm80.cuh
jiazhihao, line 1: Any reason you changed `cp.async.cg` to `cp.async.ca`?

jiazhihao: Other changes look good to me.

izhuhaoran: > Any reason you changed `cp.async.cg` to `cp.async.ca`?

According to the [PTX documentation](https://docs.nvidia.com/cuda/parallel-thread-execution/index.html?highlight=cp%2520async#data-movement-and-conversion-instructions-cp-async), `cp.async.ca` supports copy sizes of 4, 8, and 16 bytes, whereas `cp.async.cg` only supports a copy size of 16 bytes.

The documentation also states:

> "The `.cg` qualifier indicates caching of data only at the global-level L2 cache and not at L1, whereas the `.ca` qualifier indicates caching of data at all cache levels, including L1."

In most cases, it seems to me that `.ca` should yield better (or at least no worse) performance than `.cg`—even if some L1 cache space is "wasted," the performance impact should be negligible. I believe the fact that Cutlass-Cute also uses `.ca` by default further supports this view.

Additionally, I ran e2e performance test on the `demo_hopper` using the Qwen3-8B model, and observed that `.ca` is slightly faster than `.cg`, although the difference is quite small.

jiazhihao: That makes sense. The PR looks great to me. Thanks a lot for contributing.

PR #579, opened 2025-11-14
ADD split KV cache for Attention kernel

xinhaoc:
**Description of changes:**
Add split kv for both ampere and hopper attention.

**Related Issues:**

Linked Issues:
- Issue #

Issues closed by this PR:
- Closes #

### DIFF
--- a/include/mirage/persistent_kernel/runtime_header.h
+++ b/include/mirage/persistent_kernel/runtime_header.h
@@ -162,7 +162,9 @@ struct FullTaskDesc {
         trigger_event(EVENT_INVALID_ID), dependent_event(EVENT_INVALID_ID) {
     task_metadata.raw_payload = ~0ull;
   }
-  FullTaskDesc() { task_metadata.raw_payload = ~0ull; }
+  FullTaskDesc() {
+    task_metadata.raw_payload = ~0ull;
+  }
   TaskType task_type;
   unsigned variant_id;
   int num_inputs, num_outputs;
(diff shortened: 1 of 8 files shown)


include/mirage/persistent_kernel/tasks/blackwell/attention_sm100.cuh
```diff
@@ -192,8 +192,8 @@ __device__ __forceinline__ void multitoken_paged_attention_sm100_task_impl(
     constexpr size_t S_O_BUFFER_SIZE =
         sizeof(float) * MMA_ITERS_M * NUM_THREADS * 64;
     constexpr size_t S_TOTAL_OFFSET = S_O_BUFFER_OFFSET + S_O_BUFFER_SIZE;
-    static_assert(S_TOTAL_OFFSET <=
-                  mirage::runtime::MAX_DYNAMIC_SHARED_MEMORY_SIZE);
+    // static_assert(S_TOTAL_OFFSET <=
```
jiazhihao: Why we removed this static assert check?

include/mirage/persistent_kernel/runtime_header.h
```diff
@@ -210,14 +225,8 @@ struct alignas(16) TaskDesc {
...
 #endif
-  union {
-    struct {
-      int request_id; // Used for paged attention
-      int head_group; // Used for paged attention hopper
-    };
-    int expert_offset;         // Used for MoE
-    size_t xfer_size_in_bytes; // Used for nvshmem
-  };
+int request_id; // Used for paged attention
```
jiazhihao: Is it possible to move request_id into the union structure? Specifically, we can use uint16_t for both request_id and kv_id, which would allow us to move request_id to the struct for attention.

include/mirage/persistent_kernel/tasks/blackwell/attention_sm100.cuh
yzhou442: It is added back

include/mirage/persistent_kernel/runtime_header.h
yzhou442: Yes, I have moved the request_id into union. One thing to notice is that we use `request_id==-1` to verify active requests, so it is set to int16_t instead of uint16_t

jiazhihao (approved):

PR #590, opened 2025-12-01
MOE linear kernel on ampere

wenqiny:
**Description of changes:**
This PR introduce the MOE linear kernel on Ampere.

**Related Issues:**

Linked Issues:
- Issue #

Issues closed by this PR:
- Closes #

### DIFF
--- a/include/mirage/persistent_kernel/tasks/hopper/silu_mul_hopper.cuh
+++ b/include/mirage/persistent_kernel/tasks/hopper/silu_mul_hopper.cuh
@@ -15,7 +15,6 @@
 #pragma once
 #include "../common/utils.cuh"
 #include "../common/worker_config.h"
-
 namespace kernel {
 
 template <typename T,


include/mirage/persistent_kernel/tasks/hopper/silu_mul_hopper.cuh
JackFram, line 1: For silu_mul, we can use Ampere's silu_mul task impl directly

include/mirage/persistent_kernel/tasks/ampere/moe_linear.cuh
JackFram, line 1: Notice that the input shape is slightly different for W13 GeMM (bs, hidden) and W2 GeMM (bs, num_topk, inter)

wenqiny: Thanks for the notification! I will update it when I'm integrating the W2 GEMM.

include/mirage/persistent_kernel/tasks/hopper/silu_mul_hopper.cuh
wenqiny: IIUC, the unit tests will cover several gens of GPUs, so we should also keep the Hopper `silu_mul` test here, right? What I do here is just left a macro which works in unit test, that's just a workaround for building the test, WDYT?

JackFram: For unit test, we can also use a separate binding file to only include necessary functions so that the compilation process would be relatively faster. For instance, under blackwell's unit test folder, we can have different folders for testing different modules. Addtionally, we might will remove the silu_mul_hopper in the future because Ampere/Hopper/Blackwell can all use the same silu_mul task impl

wenqiny: I see, thank you for the clarification! I tried to define a macro `MIRAGE_GRACE_HOPPER` which is a workaround the unit test build issue.

The reason that I don't remove `silu_mul_hopper` is that I found there is some code which invokes the `silu_mul_hopper`, but I don't have an Hopper cards to test it after we replacing it with the Ampere implementation. I guess we could try to remove it later, because the `silu_mul_hopper` is not very related to this PR, WDYT?

https://github.com/mirage-project/mirage/blob/706151b6e20f1e282ae2b95a5031f31c2cc43bb6/src/kernel/task_register.cc#L1562-L1566

PR #614, opened 2026-01-29
NVSHMEM tile-based allreduce

### LINKED ISSUE #606: [Feature Request] - Multicast for Multi-GPU
### Feature Description

Using multicast object in CUDA for allreduce collective communications.

ZephyrZhao247:
**Description of changes:**

Add nvshmem tile allreduce

- [x]  Query device support. Especially nvlink for hopper and blackwell. If there is no NVLINK, fallback to old solution.
- [x]  ~~use nvshmem_align to allocate memory~~
- [x]  ~~Prepare Tensor structs on Host~~
- [x]  Change python `persistent_kernel.py`, to call different allreduce tasks based on architecture.
- [x]  Change task registration, stop hard coding the nvshmem copy task
- [x]  split old and new tasks.
- [x]  Add new task for nvshmem tile collectives, must be constrained inside macros NVSHMEM
    - [x]  ~~How to pass Tensor structs to Kernel?~~
        - [x]  ~~Add a new field in runtime config. `Tensor nvshmem_tile_tensors[NUM_WORKERS]`.~~
        - [x]  ~~It only stores the shape and stride info. The base ptr will be updated using the kernel paramter “input_ptr”.~~
    - [x]  Or maybe just construct at runtime? Because it’s a lightweight struct. The biggest concern here  is the register pressure.

New Dependency:

- cuda-python

# Performance

On B200: p1d512 per token latency

| Num B200s | old allgather+reduce(ms) | new multimem-based(ms) |
| --- | --- | --- |
| 1 | 4.9 | 4.9 |
| 2 | 5.6 | 4.5 |
| 4 | 5.8 | 4.0 |
| 8 | - | 3.4 |

**There is one issue worth noting**: Due to the introduction of nvshmem tile operations, we must setup NVSHMEM teams. Each SM worker requires 1 dedicated communication team. This allocation would take tens of seconds for 2 GPUs (For 8 GPUs, this would even take near 180s). This is a huge **unavoidable** setup overhead, and so is the destroy overhead.

The cost has been reduced to (4GPUs, 15s) (8GPUs, 80s).

# Performance tuning tasks

- [x] Scalability
- [x] Setup and teardown overhead. This is unavoidable.
- [x] ~~Tune parameters~~. There is no auto-tuning of parameters right now.
- [x] ~~Try different APIs. ~~ We should not use warp or warpgroup level APIs. For that, we need to create a huge amount of NVSHMEM teams, which is too expensive.
- [x] Don't transfer inactive tokens
- [x] During setup, we will see each process allocating memory on all other GPUs. Find out why.
- [x] compare with vllm and sglang
- [x] Test with a brand new env to check dependency versions.

image

The above is the benchamrk results generated using vllm's benchmark scripts. Results show that multimem-based methods perform the best under large data size, while giving fairly good perf with small data size. In general, this multimem method could serve as a general backend for allreduce.

**Related Issues:**

Linked Issues:
- Issue #606 #616 

Issues closed by this PR:
- Closes #606 #616

### DIFF
--- a/python/mirage/mpk/multigpu.py
+++ b/python/mirage/mpk/multigpu.py
@@ -74,15 +74,26 @@ def get_collective_capabilities(num_devices: int, device_id: int = 0) -> Collect
 def allocate_nvshmem_teams(mpk, num: int):
     # We should set NVSHMEM_MAX_TEAMS environment variable
     nvshmem_max_teams = os.environ.get("NVSHMEM_MAX_TEAMS", None)
-    target_num_teams = num * 20 + 32
+    # We add 32 extra teams as a buffer to avoid hitting the limit in some edge cases
+    # 6 is the number of builtin teams in NVSHMEM.
+    max_num_teams = num + 6 + 32
     if nvshmem_max_teams is None:
-        os.environ["NVSHMEM_MAX_TEAMS"] = str(target_num_teams)
+        os.environ["NVSHMEM_MAX_TEAMS"] = str(max_num_teams)
     else:
         existing_max_teams = int(nvshmem_max_teams)
-        if existing_max_teams < num:
-            os.environ["NVSHMEM_MAX_TEAMS"] = str(target_num_teams)
+        if existing_max_teams < max_num_teams:
+            os.environ["NVSHMEM_MAX_TEAMS"] = str(max_num_teams)
     mpk.allocate_nvshmem_teams = num
     # print(f"Set NVSHMEM_MAX_TEAMS={os.environ['NVSHMEM_MAX_TEAMS']}")
+
+    # We should also set NVSHMEM_MAX_CTAS environment variable to avoid creating
+    # too many duplicate teams (team_dups[] in NVSHMEM).
+    # This env is not documented in the NVSHMEM documentation.
+    # One should look at the source code of NVSHMEM to find it.
+    nvshmem_max_ctas = os.environ.get("NVSHMEM_MAX_CTAS", None)
+    os.environ["NVSHMEM_MAX_CTAS"] = str(1)
+    if nvshmem_max_ctas is not None:
+        print(f"MPK: forcing env NVSHMEM_MAX_CTAS=1.")
 
 # ============================================================================
 # Strategy Pattern: Base Classes for Collective Implementations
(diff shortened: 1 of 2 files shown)


include/mirage/persistent_kernel/tasks/hopper/allreduce.cuh
```diff
@@ -0,0 +1,141 @@
...
+  // Output stride is the same as hidden size
+  using c_hidden = ConstInt<OUTPUT_STRIDE>;
+  using c_output = ConstInt<OUTPUT_SIZE>;
+  using c_batch = ConstInt<BATCH_SIZE>;
+  using c_1 = ConstInt<1>;
+
+  constexpr int WARP_SIZE = 32;
+  int const warp_id = threadIdx.x / WARP_SIZE;
+  int const num_warps = (blockDim.x + WARP_SIZE - 1) / WARP_SIZE;
+  int const tokens_per_warp = active_tokens + num_warps - 1;
```
xinhaoc: based on the name, should this be`tokens_per_warp = (active_tokens + num_warps - 1) / num_warps` ?

python/mirage/mpk/multigpu.py
```diff
@@ -0,0 +1,287 @@
...
+    """Get or create the CollectiveCapabilities instance."""
+    global collective_capabilities
+    if collective_capabilities is None:
+        collective_capabilities = CollectiveCapabilities(num_devices, device_id)
+    return collective_capabilities
+
+def allocate_nvshmem_teams(mpk, num: int):
+    # We should set NVSHMEM_MAX_TEAMS environment variable
+    nvshmem_max_teams = os.environ.get("NVSHMEM_MAX_TEAMS", None)
+    target_num_teams = num * 20 + 32
```
xinhaoc: not sure how this `target_num_teams` compute, could you leave a comment here? And looks like this number is not used for allocation in line 84.

xinhaoc:
The PR looks great! thanks for the commit, left two comments

include/mirage/persistent_kernel/tasks/hopper/allreduce.cuh
ZephyrZhao247: You are right. Actually, this function will never be invoked. I will remove it

ZephyrZhao247: removed

python/mirage/mpk/multigpu.py
ZephyrZhao247: The number is precisely controlled now.

Issue #622, opened 2026-02-16
[Bug] - Packaging infrastructure broken - native libs not bundled in non-editable builds, paths hardcoded to source tree


Marco-Christiani:
### Bug Description

`mirage` only works as an editable install (`pip install -e .`). A standard `pip install .` fails before it even builds because `setup.py` runs cargo/cmake at module evaluation time. Even if you work around the build, the resulting package fails on import because native shared libraries are not bundled and both `__init__.py` and the Cython extension's RPATH rely on path assumptions that only hold inside the source tree.

This prevents all non-editable usage: `pip install .`, `pip install mirage` (if published to an index), and downstream packaging (Nix, conda, multi-stage container builds).

There are four root causes:

**1. `setup.py` runs cargo + cmake at module evaluation time.**

https://github.com/mirage-project/mirage/blob/95cc6106402e07e69000109fb564ccb59de77079/setup.py#L145-L232

`setup.py` executes cargo and cmake unconditionally at the top level, outside any build command class. This means any operation that evaluates `setup.py` - including pip's `get_requires_for_build_wheel` metadata hook - triggers a full native build attempt. Since build isolation is pip's default behavior, every standard `pip install .` fails immediately because cargo is not a Python package and cannot be installed into pip's isolated build environment. There is no way to skip these steps or supply pre-built artifacts, which also blocks downstream packagers (Nix, conda, Debian, etc.) that build native dependencies separately.

**2. `__init__.py` navigates outside the package to find `.so` files.**

https://github.com/mirage-project/mirage/blob/95cc6106402e07e69000109fb564ccb59de77079/python/mirage/__init__.py#L15-L20

The code walks up from `__file__` to the repo root, then into `build/`. This only works for editable installs where `__file__` is inside the source checkout. After a non-editable install, `__file__` is under `site-packages/mirage/` and the `../..` traversal resolves to a nonexistent directory.

**3. Native shared libraries are not included in the installed package.**

The build produces `libabstract_subexpr.so` and `libformal_verifier.so` but leaves them in `build/`. `setup.py` does not declare them in `package_data` or copy them into the package tree. The only `.so` in the installed package is the Cython extension:

```
$ find /tmp/test-env/lib/python3.10/site-packages/mirage/ -name "*.so"
/tmp/test-env/lib/python3.10/site-packages/mirage/core.cpython-310-x86_64-linux-gnu.so
```

No `libabstract_subexpr.so` or `libformal_verifier.so` anywhere in the installed tree.

[...]

PR #623, opened 2026-02-17
fix: add lib/stubs to CUDA library search paths in setup.py

dezash123:
**Description of changes:**
  - Fix missing `lib/stubs` in CUDA library search paths in `setup.py`, which caused `-lcuda` linker failure on systems where `libcuda.so` is under `lib/stubs/` rather than `lib64/stubs/` (e.g. my NixOS)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -42,6 +42,7 @@ else:
 cuda_include_dir = os.path.join(cuda_home, "include")
 cuda_library_dirs = [
     os.path.join(cuda_home, "lib"),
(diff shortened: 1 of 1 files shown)


Marco-Christiani:
@dezash123 I am adding proper packaging support for NixOS, you will run into many other issues trying to run this project on NixOS right now. packaging itself will likely be out of tree, support will be in tree, stay tuned though as this may change.
Related: #622

PR #626, opened 2026-02-20
[Bug Fix] Fix linear task dispatch for non-base GPU archs & Update Z3 version

xinhaoc:
**Description of changes:**

**Related Issues:**

Linked Issues:
- Issue #

Issues closed by this PR:
- Closes #

### DIFF
--- a/.github/workflows/helpers/install_dependencies.sh
+++ b/.github/workflows/helpers/install_dependencies.sh
@@ -6,8 +6,8 @@ sudo apt install -y software-properties-common lsb-release wget python3-pip g++ 
 # Install Z3
 sudo apt-get install -y libz3-4 libz3-dev
 
-# Make sure Z3 lib is found (enforces Z3 version 4.15)
-sudo ln -s /usr/lib/x86_64-linux-gnu/libz3.so /usr/lib/libz3.so.4.15 || true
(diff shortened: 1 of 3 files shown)


python/mirage/mpk/persistent_kernel.py
```diff
@@ -972,15 +972,15 @@ def linear_layer(
         tb_graph.new_input(output, (1, -1, -1), -1, True)
         self.kn_graph.customized([input, weight, output], tb_graph)
 
-        if self.target_cc == 100:
+        if self.target_cc >= 100:
```
JackFram: Maybe need to be <120 ? Not sure if the same kernel impl works for SM120 or not, I think SM120 doesn't have tensor memory?

python/mirage/mpk/persistent_kernel.py
```diff
@@ -1011,15 +1011,15 @@ def linear_with_residual_layer(
         if self.world_size > 1 and self.mpi_rank != 0:
             enable_residual = 0
         params.append(enable_residual)
-        if self.target_cc == 100:
+        if self.target_cc >= 100:
             self.kn_graph.register_task(tb_graph, "linear_with_residual_sm100", params)
```
JackFram, line 1015: Same concern

PR #664, opened 2026-04-20
Fix SM100 MoE top-k routing edge cases

JerryLiang49:
This fixes two SM100 MoE top-k routing edge cases that caused decode instability in the 256-expert path.

Summary:
- fix active-expert-count initialization for the full-width expert configuration
- fix the subgroup shuffle-mask logic for the final live row group
- add regression coverage for the larger expert-count configuration

Impact:
- removes the first confirmed NaN source observed in Mirage decode for this model/configuration
- short decode repro now completes successfully instead of failing in the MoE routing path

### DIFF
--- a/include/mirage/persistent_kernel/tasks/blackwell/topk_softmax_sm100.cuh
+++ b/include/mirage/persistent_kernel/tasks/blackwell/topk_softmax_sm100.cuh
@@ -95,7 +95,9 @@ __device__ __forceinline__ void topk_softmax_task_impl(
       mpk_active_expert_ids[expert - start_expert] = -1;
     }
   }
-  if (threadIdx.x == NUM_EXPERTS && mpk_active_expert_ids != nullptr) {
+  // Thread 0 always exists, unlike threadIdx.x == NUM_EXPERTS when
+  // NUM_EXPERTS == blockDim.x == 256.
+  if (threadIdx.x == 0 && mpk_active_expert_ids != nullptr) {
     mpk_active_expert_ids[NUM_EXPERTS] = 0;
   }
   __syncthreads();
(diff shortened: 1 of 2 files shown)


include/mirage/persistent_kernel/tasks/blackwell/topk_softmax_sm100.cuh
```diff
@@ -95,7 +95,9 @@ __device__ __forceinline__ void topk_softmax_task_impl(
       mpk_active_expert_ids[expert - start_expert] = -1;
     }
   }
-  if (threadIdx.x == NUM_EXPERTS && mpk_active_expert_ids != nullptr) {
+  // Thread 0 always exists, unlike threadIdx.x == NUM_EXPERTS when
+  // NUM_EXPERTS == blockDim.x == 256.
```
ZephyrZhao247, line 99: 1. Why do we need to handle NUM_EXPERTS=256? I think qwen3 30B has only 128 experts

include/mirage/persistent_kernel/tasks/blackwell/topk_softmax_sm100.cuh
```diff
@@ -140,9 +142,16 @@ __device__ __forceinline__ void topk_softmax_task_impl(
...
+  uint32_t warp_mask = 0xffffffffu;
+  if constexpr (THREADS_PER_ROW != WARP_SIZE) {
+    constexpr uint32_t subgroup_mask = (1u << THREADS_PER_ROW) - 1u;
+    // The final warp can contain a single live sub-group when num_rows is odd.
+    // Restrict the shuffle mask to that sub-group instead of hard-coding the
+    // lower 16 lanes.
+    if ((num_rows % ROWS_PER_WARP) != 0 && thread_row == num_rows - 1) {
+      warp_mask = subgroup_mask << (thread_row_in_warp * THREADS_PER_ROW);
+    }
+  }
```
ZephyrZhao247, line 154: This seems like a generalization of the old logic. But for qwen3 30B, NUM_EXPERS=128 and VPT=8, the logic should be exactly the same.

So I think this is not the root cause of the qwen3 bug.

ZephyrZhao247: But I think this generalization is good and could be merged.

include/mirage/persistent_kernel/tasks/blackwell/topk_softmax_sm100.cuh
JerryLiang49: Yes, for Qwen3-30B-A3B specifically, I agree it should only use 128 experts.

I included the 256 case because this kernel/wrapper already supports both 128 and 256 expert specializations, so I think the "threadIdx.x == NUM_EXPERTS" is a latent correctness bug for the 256-expert specialization.

This is just a more general kernel fix rather than the actual Qwen3-30B issue itself.

include/mirage/persistent_kernel/tasks/blackwell/topk_softmax_sm100.cuh
JerryLiang49: I agree. For the Qwen3-30B setting (NUM_EXPERTS=12 and VPT=8), this should reduce to the same behavior as the old logic.

I think we can also treat this as a general robustness improvement rather than the actual explanation for the Qwen3 bug, and I’ll keep debugging the Qwen3 multi-GPU issue separately.

ZephyrZhao247 (approved):
LGTM, all changes are local.

PR #667, opened 2026-04-23
DeepSeek V3 on B200: FP8 + MLA TP + MLA prefill + MTP

bill810975:
# DeepSeek V3 on B200: FP8 + MLA TP + MLA prefill + MTP

TP support and bugfix for the previous pr#656

This PR brings the DeepSeek V3 FP8 inference pipeline up on B200 (SM100a) inside
MPK, adds MLA tensor parallelism (TP2/TP4/TP8), MLA chunked prefill, and MTP
speculative decoding. Demo + persistent-kernel infrastructure are cleaned up so
only the proven-working configuration ships.

### 1. DeepSeek V3 end-to-end on B200
- FP8 MoE group GEMM on sm_100a (`fp8_group_gemm_sm100.cuh`) with 4-warp
  specialization (DMA / MMA / Scale / Epilogue) using tcgen05 + UTCCP for
  scale factors.
- MLA decode + reduce kernels (`mla_decode_sm100`, `mla_reduce_sm100`),
  MLA chunked prefill (`mla_prefill_sm100`), MLA absorbed-Q weight layout
  (`q_b_proj` is `[H*576, q_lora]`, decode consumes fused, prefill splits).
- Dense FP8 linear (`linear_fp8_sm100`) + fused residual variant
  (`linear_fp8_with_residual_sm100`) matching vLLM's fused residual semantics.
- DeepSeek V3 builder (`python/mirage/mpk/models/deepseek_v3/builder.py`) wiring
  61 MoE layers + MLA + MTP layer + shared/routed experts + LM head + AllReduce.
- Demo (`demo/deepseek_v3/demo.py`) with `--layers` selective weight loading
  (HuggingFace index-based) so you can run a prefix like `--layers 0-10` on a
  single GPU without paying the full 61-layer weight load.

### 2. MLA TP (2 / 4 / 8)
- `mla_mtp_decode_tp{2,4,8}_sm100.cuh` — MLA TP decode kernels that issue
  `tcgen05.alloc` only from warp 0 and use a named barrier `bar.sync 1, 128`
  for the active-thread sync. This is required: CuTe's TMEM allocator demands
  the same warp does alloc + dealloc, and Independent Thread Scheduling
  otherwise causes warp drift.
- TP sharding rules live in two places (`builder.py` SHARD_RULES + demo
  state-dict conversion pass). Documented in `CLAUDE.md`.

### 3. MLA chunked prefill (default-on)
- `mla_prefill_sm100` dispatched automatically when `Q_LEN >= 32`
  (prefill routing in `builder.py`). Uses 576-d fused head; splits `q_b` into
  `q_b_nope[512]` + `q_b_pe[64]` in demo.py's Phase-2 absorption.
- Prefill kernel uses the `lp_ - fp_ - 1` page-indptr formula inside the
  task-register snippet so chunked prefill works with a statically-compiled
  task graph (see `src/kernel/task_register.cc`).

### 4. MTP speculative decoding (vLLM-aligned)
- Demo flag: `--mtp N` (int, 0-3; 0 = disabled, N > 0 = draft length).
- Prefill step for the MTP layer uses ground-truth prompt tokens shifted by
  one (matches `vllm/v1/spec_decode/eagle.py` L666-669), not argmax of the
  main model's logits. New task: `TASK_MTP_BUILD_EMBED_INPUT = 294`, kernel
  `mtp_build_embed_input_kernel` in `mtp_token_ops.cuh`.
- Draft→verify loop with strict / probabilistic / synthetic rejection sampling.
- MTP decoder uses the fused `linear_fp8_with_residual_sm100` kernel for
  its residual add (previous revision used a separate elementwise_add followed
  by AllReduce, which caused an `illegal memory access` on TP>=2 due to a
  double-counted NVSHMEM `sync_counter`).

### 5. rdc=true default restored on Blackwell
- Re-verified on CUDA 13.2 + NVSHMEM 3.6.5: the 2026-04 "rdc=true hangs
  sm_100a" symptom is gone. The previous register-inflation issue (166 → 255
  regs from pulling in `libnvshmem_device.a`) does not reproduce on the
  current toolchain.
- Blackwell now uses the standard NVSHMEM link path (`-rdc=true
  -lnvshmem_device`). The old self-contained-allreduce workaround
  (`-DNVSHMEM_NO_DEVICE_LIB`, hand-rolled `__managed__
  nvshmemi_device_state_d`, `nvshmemid_hostlib_init_attr` callback,
  `ar x collective_launch.cpp.o`, host-side stubs) is kept behind
  `MPK_RDC_FALSE=1` as an escape hatch for regression isolation.
- Verification (all PASS on 2026-04-22): TP=2 9.645 ms/tok; TP=4 mbt=1 spec=1
  9.280 ms/tok; TP=4 mbt=1 spec=3 13.205 ms/tok; TP=4 mbt=64 28.614 ms/tok.
  rdc=true is slightly faster than the old rdc=false workaround.

## How to run

See `demo/deepseek_v3/readme.md` for the full env setup (NVSHMEM, MPI paths).


include/mirage/persistent_kernel/tasks/blackwell/task_header.cuh
```diff
@@ -16,10 +16,13 @@
...
 #include "tasks/hopper/rotary_embedding_hopper.cuh"
 #include "tasks/hopper/silu_mul_hopper.cuh"
-#ifdef USE_NVSHMEM
+#if defined(USE_NVSHMEM) && !defined(MIRAGE_GRACE_BLACKWELL)
 #include "tasks/hopper/allreduce.cuh"
-#endif // USE_NVSHMEM
+#endif
 // Blackwell task impls
+#if defined(USE_NVSHMEM) && defined(MIRAGE_GRACE_BLACKWELL)
+#include "allreduce.cuh"
```
jiazhihao: Can we move the allreduce.cuh to `task/blackwell/allreduce.cuh`

include/mirage/persistent_kernel/persistent_kernel.cuh
jiazhihao, line 1: Why MPK_ENABLE_VERBOSE marco'ed code is removed?

jiazhihao (changes requested):

include/mirage/persistent_kernel/tasks/blackwell/task_header.cuh
bill810975: done.

include/mirage/persistent_kernel/persistent_kernel.cuh
bill810975: Sorry for this change, it's accidentally removed by coding agent while told to clean the debug flags.

This flag has been added back in new commits.

jiazhihao (approved):

PR #672, opened 2026-04-29
[Model][Kernel][Core] dpskv3 & related kernels & tensor view

ZephyrZhao247:
## Summary

This PR adds **DeepSeek V3 on Blackwell SM100** (FP8 MoE + MLA TP + chunked prefill + MTP speculative decoding) to MPK, plus a new **Tensor View API** in the IR. Across ~200 files, +30k/-2.8k lines.

### 1. Tensor View API (`mpk.view`, `mpk.narrow`, `mpk.split`)

Before this PR, every DTensor owned the storage it pointed to: dependencies
between layers were tracked by tensor GUID. While for some scenarios, like fused_linear, we might want to pass different sub-tensors to different consequent layers, which is supported by the existing tensor APIs. So we need this tensor view functionality.

A DTensor now carries three additional fields:

- `base_guid` — GUID of the root storage tensor. `0` means this tensor *is*
  storage; nonzero means it is a view of the tensor with that GUID.
- `view_offset` — byte offset from the storage tensor's base pointer.
- `stride[MAX_TENSOR_DIMS]` — per-element strides along each dim. For
  storage tensors these come from the user's input. For views they are
  computed at construction so that codegen can address the view directly
  without walking back to the parent.

Three Python methods on `PersistentKernel` produce views. They are the only
public entry points for creating non-storage DTensors:

- `view(parent, new_shape)` — reshape (same total element count, same
  byte range), e.g. `[B, H*D] → [B, H, D]`.
- `narrow(parent, dim, start, length)` — contiguous slice along one
  dimension. Most common in DSv3 (carving Q_NOPE and Q_PE out of a fused
  QKV output).
- `split(parent, n, dim)` — convenience wrapper producing N equal-sized
  sibling narrows along one dim.

#### 1.1 Dependency analysis

Every read/write of a DTensor is keyed by `resolve_base_guid()`, not by
`guid`. This is what lets the analyzer notice that two layers writing
different views of the same parent are actually writing the same storage.

Each writer is recorded with an N-dimensional bounding box describing the
sub-region it actually writes (derived from `view_offset` + `stride` +
`dim[]`). This is used to detect overlap for virtual tensors. For example, if tensor A is split into A1 and A2, when a kernel M write A1, the kernels that depend on A2 should not regard kernel M as a predecessor, which is a false dependency. We need specific bounding box to detect the exact denpendency.

A writer→reader edge is marked `is_barrier_edge` when at least one
endpoint is a view of the shared parent. For barrier edges, we enforce a single event for synchronization between layers. This is because fine-granularity dependency is very complicated in this case, since the subsequent layers may depend on data produced by part of the tasks from the previous layer.

### 2. DeepSeek V3 model on Blackwell SM100

End-to-end DSv3 demo at `demo/deepseek_v3/demo.py` with TP up to 8 and EP support for routed experts.

- **Multi-Latent Attention (MLA)**:
  - Chunked prefill (`mla_prefill_tp8_chunked_sm100`) with split-K reduction.
  - Per-TP-size decode kernels (`mla_mtp_decode_tp{2,4,8}_sm100`).
  - Unified prefill/decode dispatcher (`mla_unified_sm100`).
  - DeepSeek RoPE specialised kernel (`deepseek_mla_rope_sm100`).
  - KV gather with split fan-out across CTAs (`mla_kv_cache_gather`).
- **BMM-style absorbed Q path** (`linear_fp8_bmm_sm100` + `assemble_q_decode_sm100`), default ON: per-head FP8 BMM replacing the absorbed q_b_proj monolith, smaller TMA traffic.
- **Speculative decoding (MTP)** kernels: `target_verify_mtp`, `mtp_token_ops` (prepare-verify, accept-commit, token-scatter, build-embed-input).

### 3. FP8 GEMM family

- **Dense FP8 GEMM**: `fp8_gemm_dense_smallm/mediumm/decode_splitk` sharing a templated common impl. Block-scale MMA with UE8M0 scales.
- **Group FP8 GEMM** (MoE experts): `fp8_group_gemm_smallm/largem` — fused B-scale MMA, 5 TMA descriptors (A, B, SFA, SFB, D).
- **swapAB FP8 linear** (`linear_fp8_swapAB_sm100`, renamed from `linear_fp8_1d2d_sm100`) for decode-shape (M≤16) where the A-side becomes the "small" operand.
- **Fused RMSNorm + per-token-group FP8 quantize** (`fused_rmsnorm_quantize_fp8_sm100`): one task replaces the bf16-rmsnorm → quantize_fp8 chain, eliminating one dispatch wave + one BF16 HBM round-trip per layer. Default ON.

### 4. MoE infrastructure

- **TopK sigmoid routing** (`topk_sigmoid_sm100`) with grouped-experts gating.
- **MoE permute/unpermute** (`moe_permute_sm100` + `moe_unpermute_sm100`) — peripheral glue for the grouped-GEMM pre-permuted-input contract.
- **Active-row skipping** for MoE silu_mul and quantize (`MPK_DSV3_ACTIVE_SKIP=1`).
- **Fused `moe_mul_sum_add`** for the unpermute+residual combine.

### 5. Runtime + compiler

- **Persistent-kernel split-vs-single launch**: `split_worker_scheduler=true` is the validated default again (verified to NOT hang on B200 despite 128+80 CTAs > 148 SMs — CUDA co-schedules over time).
- **Cycle-detection + residual-stripping** (`annotated_graph.cc` steps b/c) and **fork/join LCM** passes (steps h/i) for the per-edge `event_dim` graph annotation.
- **Profiler**: `python/mirage/mpk/profiler_persistent.py` with CSV export, paired BEGIN/END detection, `EventType.kInstant` support. `scripts/mpk_perfetto_timeline.py` parses traces.
- **Test mode redesign**: `pk()` is now callable; the C++ side runs through the standard `prepare_next_batch` and terminates naturally. No need for an explicit `is_test_mode` parameter. *(This supersedes `mpk` PR #673; see "Merge notes" below.)*

### 6. Hygiene + safety

- ~50 commits of post-review hygiene: removed pre-View-API `IN_OFFSET`/`OUT_OFFSET`/`K_PE_OFFSET` template params (always 0 in production); dropped NULL-result perf-experiment env hooks (C12, C13, C17, C18); stripped dead `MPK_DEBUG_*`/`MIRAGE_DEBUG_*`/`NVSHMEM_NO_DEVICE_LIB` branches.
- New static asserts for FP8 group-tile alignment, MoE permute overflow trap, FP8 GEMM expert-id uniformity.
- `Graph::view_storage` `unique_ptr` ownership for view DTensors (fixes a per-compile leak).

## Merge notes

`mpk` is currently `524c4e0b` ("fix: recover test mode" #673) which adds an `is_test_mode` parameter to `init_persistent_kernel`. This branch solves the same problem differently — the test-mode redesign here removes the need for that explicit parameter — so a textual merge would compile-fail (the merged `init_func` references an undeclared `is_test_mode`). **#673 is semantically subsumed by this PR**; recommend merging via the GitHub "merge" button (which will pull in the parameter and we should follow up with a small cleanup), or land this PR first and revert/remove #673's parameter as a follow-up commit on `mpk`.

## Test plan

- [x] `python demo/qwen3/demo.py --use-mirage --model /raid/catalyst/models/Qwen3-8B --max-num-batched-requests 1` — coherent output, 4.34 ms/token.
- [x] `mpirun -np 4 python demo/deepseek_v3/demo.py --use-mirage --layers 0-8 --prompt-length 140 --max-num-batched-tokens 128 --max-seq-length 160 --model-path /raid/catalyst/models/DeepSeek-V3` — 85,381 tasks, all 4 ranks finish, 1.69 ms/token.
- [x] `test_view_corner_cases.py` (660 lines) exercises 8 structural view patterns with both numerical assertions and AnnotatedGraph dump verification (edges, barriers, fork/join groups).
- [x] All per-layer test-mode tests under `tests/runtime_python/blackwell/sm100_*/test_*_testmode.py`.


xinhaoc:
Hi @ZephyrZhao247 , could you resolve the conflicts with the mpk branch?

PR #683, opened 2026-05-08
dpsk v3 pytorch reference, bugfix, merge new kernels

bill810975:
done:
1. build pytorch reference to check the correctness.
2. merged PR674 chunked MLA prefill, fp8 GEMM.
3. bugfix for MTP.
4. bugfix for chunk prefill path.
5. dual path: early return or change the runtime?        now: early return.
6. bugfix for early return layer
7. fused tensor

todo:
1. standalone test and merge fp8 group gemm kernel. (ongoing)
2. do the kernel to kernel comparison, provide the kernel name to @xinhaoc 
3. merge bmm.
4. 3us linear after rmsnorm?
5. replace identity layer with dummy layer
6. check why decode mla performance downgraded.
7. mla decode kernel support paged cache
8. optimize grid dim
9.


ZephyrZhao247 (changes requested):
1. Do we really need the files in the folder "scratch"? If it's just for development, don't commit them.
2. And also "scripts"?
3. The dpskv3_reference is not complete, and rely on external python scritps. Let's remove them too. We will figure out a better way to do the reference comparison later.
4. There are so many debug related Macros, control logic, flags, envs. Let's clean them up.
5. Let's revert the fused linear for q_a_proj, and the related modifications to the RMSNorm and Rope kernels. We will use tensor view approach to solve this.
6. After cleanup, give a summary of changes in this PR, based on commit history.

bill810975:
The cleanup version is in #687

PR #692, opened 2026-05-17
[MPK] Support EAGLE3 Speculative Decoding for Qwen3

Risc-lt:
## Description

Adds end-to-end EAGLE3 speculative decoding to MPK for Qwen3-30B-A3B (MoE, 48-layer), single-GPU B200 (sm_100a). The target forward, the K-step EAGLE3 draft loop, and strict verify are all built inside one persistent megakernel; the draft loop reuses the existing `paged_attention_sm100` kernel via two new compile-time template params (`Q_LEN_OVERRIDE`, `TAIL_OFFSET`) instead of adding new attention kernels.

The output is **token-for-token correct** vs a greedy PyTorch Qwen3-30B-A3B reference for K=1..5 (K=1..4 exact for the first 50 tokens; K=5 differs only at a single genuine bf16 argmax tie). Correctness rests on fixing an in-place MMA accumulator bug (`mma.cuh`: the accumulator is now a single read-write `+f` operand, so unrolled MMA tiles no longer cross-contaminate); the earlier `m>0` MMA-skip workaround is removed and all attention tiles are computed.

Largest supported draft length is K=5 (mbt=6): the baked SM100 attention `MAX_TOKENS` is 6, fitting the 207 KB B200 offline shared-memory budget. K=6 (mbt=7) is shared-memory-bound and left for a follow-up.

## Results

K=0 = no-spec baseline. All runs: B200 single-GPU offline, Qwen3-30B-A3B, prompt-length 39, generate-length ~517 (`--ignore-eos`), per-token latency over the full prefill+decode run.

| K | mbt | per-token latency | vs no-spec |
|---|-----|-------------------|------------|
| 0 (no-spec) | 1 | 6.77 ms | ref |
| 1 | 2 | 5.16 ms | **+31%** ✓ |
| 2 | 3 | 5.94 ms | **+14%** ✓ |
| 3 | 4 | 6.57 ms | **+3%** ✓ |
| 4 | 5 | 7.92 ms | −15% |
| 5 | 6 | 8.41 ms | −20% |

K=1 is the sweet spot. For K≥4 the per-iter draft cost outweighs the marginal extra accepted tokens (accept rate does not keep up), so throughput drops below no-spec. (These numbers are post-fix — with all MMA tiles computed, not the earlier m>0-skip workaround.)

## Per-iter task order

```
Iter N:
  ┌─────────────────────────────────────────────────────────────┐
  │ Step 1: Target forward (48 layers)                          │
  │   Each layer: paged_attention_sm100 (default params)        │
  │   Layer 2, 24, 45: tee hidden → aux_h0, aux_h1, aux_h2      │
  │   Final norm + lm_head → argmax_reduce → target_argmax      │
  │                                         (mbt slots)         │
  └─────────────────────────────────────────────────────────────┘
                            │
                            ▼
  ┌─────────────────────────────────────────────────────────────┐
  │ Step 2: Verify + commit                                    │
  │   mtp_verify_strict (prev iter's drafts vs target_argmax)  │
  │     → accepted_count                                        │
  │   eagle3_commit (write accepted prefix + new drafts into    │
  │     tokens, step += accepted_count, snapshot drafts_prev)  │
  └─────────────────────────────────────────────────────────────┘
                            │
                            ▼
  ┌─────────────────────────────────────────────────────────────┐
  │ Step 3: Draft loop (K steps)                                │
  │                                                             │
  │   for step in range(K):                                     │
  │     input_token = target_argmax (step 0)                    │
  │                 = prev step's d2t-remapped token (step > 0) │
  │     embed → enorm + hnorm → concat(2H) → fused_qkv          │
  │                                                             │
  │     paged_attention_sm100:                                  │
  │       K=1: default (writes [step, step+mbt))                │
  │       K>1: Q_LEN_OVERRIDE=1, TAIL_OFFSET=K-step             │
  │            → writes one position; attends [0, .)            │
  │            → (sees prior steps' K/V)                        │
  │                                                             │
  │     o_proj + residual → post_norm → MLP → draft_hidden      │
  │     final_norm → lm_head(32256) → argmax_partial+reduce     │
  │     d2t_remap → target_token                                │
  │     scatter target_token → all_draft_ids[:, step]          │
  └─────────────────────────────────────────────────────────────┘
```

Each draft step k+1's `embed` input depends on step k's `target_token` (a DTensor). The task graph's release/acquire fence on `target_token` orders step k → step k+1, including step k's K/V cache writes, so step k+1's attention sees them naturally — no explicit dummy tokens needed.

image

## Test

- [x] **Output correctness vs greedy PyTorch** — the EAGLE3 megakernel output is   compared token-for-token against a standalone greedy (argmax) PyTorch   `Qwen3MoeForCausalLM` decode on the same prompt. Strict rejection sampling makes   the accepted-token stream equal the greedy target decode independent of K, so   greedy HF decode is the exact oracle. Result (first 50 generated tokens, B200):   K=1, K=2, K=3, K=4 exact; K=5 matches except one genuine bf16 argmax tie (two candidates with equal bf16 logits, broken the other way by the megakernel's reduction order — not an error). The correctness harness (reference generator, `--save-tokens` flag, pytest comparison, runner) lives on branch `spec-eagle3-correctness-test`, intentionally not part of this PR.
- [x] **Latency** — K=0..5 benchmarked on B200 (see Results).
- [x] **Smoke test** — [tests/runtime_python/test_eagle3_smoke.py]（tests/runtime_python/test_eagle3_smoke.py) exercises EAGLE3 task-graph generation without a GPU launch.
- [ ] **K=6** — shared-memory-bound on B200 (mbt=7 → MAX_TOKENS=7 → ~228 KB > 207 KB budget); needs an attention-shmem reduction.
- [ ] **DeepSeek V3 MTP demo regression** — shares `paged_attention_sm100`; should be re-run since the MMA accumulator helper (`mma.cuh`) changed (the fix is safe: all ~22 callers pass `C == D`, verified).

## Command

```bash
python \
  demo/qwen3/demo_30B_A3B_eagle3.py \
  --use-mirage --eagle3 \
  --model Qwen/Qwen3-30B-A3B-Instruct-2507 \
  --eagle3-draft-path lmsys/SGLang-EAGLE3-Qwen3-30B-A3B-Instruct-2507-SpecForge-Nex \
  --max-num-batched-tokens 2 \
  --num-draft-steps 1 \
  --max-seq-length 556
```


include/mirage/persistent_kernel/tasks/blackwell/attention_sm100.cuh
```diff
@@ -40,7 +40,9 @@ template <typename T,
           int HEAD_DIM,
           int MAX_SEQ_LEN,
           int PAGE_SIZE,
-          int MAX_TOKENS = 8>
+          int MAX_TOKENS = 3,
```
xinhaoc: why we can't keep this MAX_TOKENS=8?

include/mirage/persistent_kernel/tasks/blackwell/attention_sm100.cuh
```diff
@@ -445,6 +455,35 @@ __device__ __forceinline__ void multitoken_paged_attention_sm100_task_impl(
...
+        // same recovery. So the m=1 MMA is corrupting m=0's x_frag_f despite
+        // them being declared in separate elements of a `float[MMA_ITERS_M][8]`.
+        //
+        // Phase 3 attempts that DID NOT help (the contamination persists at
+        // the SASS level, with NVCC producing functionally identical kernel
+        // binaries):
+        //   - `+f` read-write constraint on MMA macro
+        //   - per-m q_frag/kt_frag inside the loop body
+        //   - `asm volatile("":::"memory")` barrier between m-iters
+        //
```
xinhaoc: i guess we need a multi token kernel? if you could provibe me the Q_LEN i can add it

include/mirage/persistent_kernel/tasks/blackwell/attention_sm100.cuh
Risc-lt: This is just an ad-hoc patch to run model on B200 with enough smem space. We could remove this patch when merging.

include/mirage/persistent_kernel/tasks/blackwell/attention_sm100.cuh
Risc-lt: Sure, the multitoken_paged_attention_sm100_task_impl seems not able to handle both the target and draft model's attention correctly at the same time. The root cause may be the conflict of different input token numbers, which are MAX_TOKENS and 1, respectively.

include/mirage/persistent_kernel/tma.cuh
```diff
@@ -1620,6 +1620,14 @@ __host__ inline void create_tma_desc_by_task(FullTaskDesc &task_desc) {
...
       break;
     }
+    // Eagle3 kernels don't need TMA. 
+    case TASK_COPY:
+    case TASK_EAGLE3_AUX_CONCAT:
+    case TASK_EAGLE3_INPUT_CONCAT:
+    case TASK_EAGLE3_D2T_REMAP:
+    case TASK_EAGLE3_COMMIT: {
+      break;
+    }
```
xinhaoc: we don't need this in `tma.cuh `?

include/mirage/persistent_kernel/runtime_header.h
```diff
@@ -143,6 +143,11 @@ enum TaskType {
   // SM100 Tasks
   TASK_SM100_TASK_BEGIN = 230, // SM100 start placeholder, not a real task
   TASK_SM100_TMA_START_TASK = 231,
+  TASK_COPY = 232,
+  TASK_EAGLE3_AUX_CONCAT = 233,
+  TASK_EAGLE3_INPUT_CONCAT = 234,
```
xinhaoc: is there an opportunity to reduce the task number(like fuse contact into single task. so we can support any shape of concact

include/mirage/persistent_kernel/tma.cuh
Risc-lt: Yeah, these newly added kernels do not use TMA but need to be dispatched here in order to pass the building assertions.

These tasks will be routed into `create_tma_desc_by_task` (the dispatch condition in runtime.cc is `231 < task_type < 256`, and `TASK_COPY=232` through `TASK_EAGLE3_COMMIT=236` all fall within this range).

On Blackwell, the build is configured with `-DMPK_ENABLE_TMA` and without `-DNDEBUG`, so assert(false) remains active. If these cases are missing, the program will abort at runtime.

include/mirage/persistent_kernel/runtime_header.h
Risc-lt: Sure, I agree. I'll replace them with one general concat task.

include/mirage/persistent_kernel/tma.cuh
xinhaoc: i see, we may need to update the range of task number

include/mirage/persistent_kernel/tasks/blackwell/attention_sm100.cuh
jiazhihao: My understanding (after taking to Letian) is that attention performs more token due to speculative decoding. Is this a prefill task or decode task?

src/kernel/graph.cc
```diff
@@ -832,6 +833,33 @@ void Graph::register_task(char const *task_type, std::vector<int> params) {
     task_config[op] =
         std::make_tuple(1, 1, TASK_MTP_BUILD_EMBED_INPUT, variant_id);
   }
+  // Eagle3 tasks
+  else if (name == "copy") {
+    int variant_id =
+        task_register->register_copy_task(customized->bgraph, params);
+    task_config[op] = std::make_tuple(1, 1, TASK_COPY, variant_id);
+  } else if (name == "eagle3_aux_concat") {
```
jiazhihao: Why do we still have aux_concat and input_concat?

src/kernel/graph.cc
```diff
@@ -832,6 +833,33 @@ void Graph::register_task(char const *task_type, std::vector<int> params) {
     task_config[op] =
         std::make_tuple(1, 1, TASK_MTP_BUILD_EMBED_INPUT, variant_id);
   }
+  // Eagle3 tasks
+  else if (name == "copy") {
```
jiazhihao, line 837: What's the purpose of copy task here?

src/kernel/task_register.cc
```diff
@@ -4393,6 +4400,105 @@ int TaskRegister::register_mtp_build_embed_input_task(
...
+
+  mirage::transpiler::CodeKeeper code;
+  code.inc_indent();
+  code.e("kernel::copy_layer_kernel<bfloat16, $, $>(", batch_size, hidden_dim);
+  code.e("    task_desc->input_ptrs[0],");   // src
+  code.e("    task_desc->output_ptrs[0]);"); // dst
+  return register_task_variant(TASK_COPY, code.to_string());
+}
+
+int TaskRegister::register_eagle3_aux_concat_task(
```
jiazhihao: aux_concat and input_concat should be merged into a more general concat task.

jiazhihao (changes requested):

src/kernel/graph.cc
Risc-lt: The copy task captures the hidden states into a dedicated aux buffer during the target model's forward pass. It's needed because the intermediate buffers are shared and reused across layers, so a captured hidden would otherwise be overwritten by the next layer. We snapshot the pre-layer hiddens at layers 2 / 24 / 45 (low/mid/high), which EAGLE3's draft forward consumes as input.

src/kernel/task_register.cc
Risc-lt: At the kernel level, both now call a single generic `concat_kernel<T, B, H, N>` with N=3 / N=2; the old `eagle3_aux_concat_kernel` / `eagle3_input_concat_kernel` are gone.

Currently these two task types (TASK_EAGLE3_AUX_CONCAT / TASK_EAGLE3_INPUT_CONCAT) are separate and only differ by the N template arg now. Yeah, I can collapse them into one concat task carrying N as a param if you'd prefer.

src/kernel/graph.cc
Risc-lt: I leave them separate to show they are two distinct steps in the EAGLE3 data flow. But in fact they share a similar kernel pattern. I can put them together. FYI, 

`aux_concat`: draft step 0 only — concatenates the 3 captured aux hiddens (3×(B,H)→(B,3H)) to feed w_fc, which projects 3H back to H.
`input_concat`: every draft step — concatenates embed + hidden (2×(B,H)→(B,2H)) to feed the draft QKV projection.

src/kernel/task_register.cc
```diff
@@ -1943,17 +1947,20 @@ int TaskRegister::register_paged_attention_sm100_task(
 
   mirage::transpiler::CodeKeeper code;
   code.inc_indent();
+  // MAX_TOKENS=3 supports Eagle3 K=2 (mbt=3 target attention).
   code.e("kernel::multitoken_paged_attention_sm100_task_impl<bfloat16, $, $, "
          "$, $, "
-         "$, $, $, $>(",
+         "$, $, $, $, 6, $, $>(",
```
jiazhihao: This change sets the number of tokens to 6 instead of the defaulty 8.

Risc-lt: Back to 8 now.

Risc-lt:
All comments solved. Please re-check it. cc @xinhaoc @jiazhihao

jiazhihao (approved):

PR #693, opened 2026-05-21
Add SM100 planned attention support

tangzhanshuo:
Implements a FlashInfer-style attention planner plus the supporting SM100 prefill kernel and end-to-end planned dispatch, replacing the per-request 1-CTA dispatch for mixed prefill/decode batches on Blackwell.

## Changes

New SM100 kernels (include/mirage/persistent_kernel/tasks/blackwell/)
- attention_prefill_sm100.cuh — wide-Q paged-attention prefill. Split-M warp layout: each of 4 warps owns NUM_Q/4 packed-Q rows and walks the full KV tile, avoiding the cross-warp m/d/o smem buffers in the decode kernel. Supports paged KV write-back, qk-norm/RoPE, and an optional KV-chunk + LSE output for split-KV.
- attention_planner_sm100.cuh — single-CTA planner. Classifies each request as prefill (packed_qo > PREFILL_THRESHOLD) or decode, tiles by P_Q_TILE/D_Q_TILE, optionally chunks prefill KV by KV_SPLIT_SIZE. Greedy min-cost bucket assignment (round-robin fast path while assigned < buckets, heap-sift afterward). Three phases: serial greedy assign on
thread 0 → prefix-sum bucket counts on thread 0 → parallel global store of worker_* arrays across all 128 threads.
- attention_planned_prefill_sm100.cuh / attention_planned_decode_sm100.cuh — per-bucket consumers that read their slice of the planner's work list and dispatch the prefill / decode impl per work item. Cheap early-exit on empty buckets.
- attention_planned_dual_sm100.cuh — fused consumer that runs both prefill and decode works for one bucket in a single CTA, halving consumer CTA count vs the split form.

Merge variants (tasks/ampere/merge_splitkv.cuh)
- merge_splitkv_persistent (mode 1) — static per-request/head/split.
- merge_splitkv_planned_persistent (mode 2) — atomic work-queue keyed on a counter in the plan buffer; iterates total_tokens × total_qo_heads rows. Both filter decode-only and direct-unsplit-prefill rows so split-KV only merges the prefill subset.

Decode kernel extension (attention_sm100.cuh)
- Added optional kv_chunk_idx / kv_chunk_size / lse_ptr / flat_kv_head_idx / flat_num_kv_heads / prefill_threshold params, plus a partial_group_offset so split-KV partials land in per-chunk output slots. Existing call sites unchanged.

Task registration (src/kernel/task_register.cc, include/mirage/kernel/task_register.h)
- Five new task variants: TASK_PAGED_ATTENTION_PREFILL_SM100 (296), TASK_ATTENTION_PLANNER_SM100 (297), TASK_PLANNED_PREFILL_ATTENTION_SM100 (298), TASK_PLANNED_DECODE_ATTENTION_SM100 (299), TASK_PLANNED_DUAL_ATTENTION_SM100 (300). Multi-GPU range moved to 350+ to accommodate.
- register_paged_attention_sm100_task gains an optional prefill_threshold parameter that emits a runtime-dispatched dual-kernel call (large packed_qo → prefill kernel, otherwise decode kernel) — the baseline against which the planner is compared.
- register_paged_attention_split_kv_merge_sm100_task gains a merge_mode param selecting legacy / persistent / planned merge.


include/mirage/persistent_kernel/tasks/blackwell/attention_planned_prefill_sm100.cuh
```diff
@@ -0,0 +1,184 @@
...
+  constexpr int PLAN_SIZE = 2 * (NUM_BUCKETS + 1) + 4 * MAX_WORKS + 2;
+  constexpr int PROFILE_STRIDE = 4 + 8 * NUM_BUCKETS;
+  int *plan_buffer_rw = const_cast<int *>(plan_buffer);
+  int *profile_base = plan_buffer_rw + PLAN_SIZE + PROFILE_SLOT * PROFILE_STRIDE;
+  unsigned long long _cons_start_ns;
+  if (threadIdx.x == 0) {
+    asm volatile("mov.u64 %0, %%globaltimer;" : "=l"(_cons_start_ns));
+    reinterpret_cast<unsigned long long *>(
+        profile_base + 4 + 4 * bucket_idx)[0] = _cons_start_ns;
+  }
```
xinhaoc: these are profiling code, we wrap it at the runtime before and after call a task. i guess we can remove this before merging the pr

include/mirage/persistent_kernel/tasks/blackwell/attention_planned_prefill_sm100.cuh
```diff
@@ -0,0 +1,184 @@
...
+  // completed their writes before stamping the end time.
+  __syncthreads();
+  __threadfence();
+  if (threadIdx.x == 0) {
+    unsigned long long _cons_end_ns;
+    asm volatile("mov.u64 %0, %%globaltimer;" : "=l"(_cons_end_ns));
+    reinterpret_cast<unsigned long long *>(
+        profile_base + 4 + 4 * bucket_idx)[1] = _cons_end_ns;
+  }
+}
```
xinhaoc: same at here

include/mirage/persistent_kernel/tasks/blackwell/attention_prefill_sm100.cuh
xinhaoc, line 1: is this kernel for GQA prefilling? Do we need to further optimize the performance of it

xinhaoc:
Hi @tangzhanshuo if i didn't misunderstand the `planning_paged_attention_layer` is called per attention layer. it should  called once per iteration and resuse for all attention kernels

tangzhanshuo:
I've removed the profiling code and modified the planning to only run once per iteration.

include/mirage/persistent_kernel/tasks/blackwell/attention_prefill_sm100.cuh
tangzhanshuo: Yeah its the prefill kernel, could be optimized further

src/kernel/task_register.cc
```diff
@@ -4678,5 +4914,389 @@ int TaskRegister::register_mla_mtp_decode_tp8_reduce_sm100_task(
...
+         page_size,
+         max_tokens);
+  code.e("    task_desc->input_ptrs[0],");   // qkv base
+  code.e("    task_desc->input_ptrs[1],");   // k_cache base
+  code.e("    task_desc->input_ptrs[2],");   // v_cache base
+  code.e("    task_desc->output_ptrs[0],");  // output base
+  // The consumer kernel takes only its own indptr (either prefill or decode).
+  code.e("    static_cast<int const*>(task_desc->input_ptrs[$]),",
+         plan_indptr_input_idx);
+  code.e("    static_cast<int const*>(task_desc->input_ptrs[9]),");   // worker_batch
```
xinhaoc: this exceed the MAX_INPUTS_PER_TASK?

xinhaoc:
BTW could you resolve the conflict? @tangzhanshuo

src/kernel/task_register.cc
tangzhanshuo: That was unused helper code for an older version of the split plan representation, removed now

tangzhanshuo:
The shared memory change was because the worker also compiling in the planned attention prefill/decode paths, leading to it taking up more static smem, therefore dynamic smem needs to be reduced.
Merging the planning into prepare_next_batch is a bit complicated, as prepare_next_batch currently runs on a single thread, and the planning uses multiple threads for parallelization optimizations. I tested the planner launched with 128 threads vs 1 thread, and it is slightly slower for single thread (like 30 vs 40 us). Do we take that and merge the single threaded planner into prepare_next_batch @xinhaoc

demo/qwen3/demo.py
```diff
@@ -556,6 +590,24 @@ def max_factor_leq_n(m: int, n: int) -> int:
                     grid_dim=(1, num_local_kv_heads, 1), #TODO: further divide across batch dim
                     block_dim=(128, 1, 1),
                 )
+            elif args.use_planned_attn:
```
jiazhihao: Instead of manually adding a planning layer, let's make it part of prepare_next_batch

include/mirage/kernel/task_register.h
jiazhihao, line 1: We will switch to attention planning for all models. Instead of registering new tasks, can we update the implementation of existing attaention taasks to make them compatiable with attention planning?

include/mirage/persistent_kernel/tasks/ampere/multitoken_paged_attention_4_16_split_kv.cuh
jiazhihao, line 1: What's the difference between kv_idx and hv_head_idx?

include/mirage/persistent_kernel/tasks/blackwell/attention_sm100.cuh
jiazhihao, line 1: What are these additional parameters for attention?

include/mirage/persistent_kernel/runtime_header.h
jiazhihao, line 1: We should discuss this change as it affects all other tasks.

jiazhihao (changes requested):

tangzhanshuo:
Everything should be cleaned up and fixed now.

PR #703, opened 2026-05-30
Update Qwen3 CI workflow: conda env, idle GPU selection, cleanup

Electronic-Waste:
## Summary

Overhaul the Qwen3 CI pipeline (`ci-tests-qwen3.yml`) for the self-hosted B200 runner, making it more robust and resource-efficient. Also restrict all workflow push triggers to the `mpk` branch to avoid duplicate CI runs on PR branches.

### `ci-tests-qwen3.yml`
- Replace `actions/setup-python@v5` with `conda-incubator/setup-miniconda@v3` for proper conda environment management
- Add `defaults: run: shell: bash -l {0}` so conda is initialized in every step
- Isolate conda environments per run using `mirage-ci-${{ github.run_id }}` to prevent cross-run pollution
- Auto-select an idle GPU via the new `find_idle_gpus.sh` script instead of hardcoding `CUDA_VISIBLE_DEVICES`
- Deactivate conda before removing the environment to avoid errors
- Clean up workspace contents between runs to prevent stale artifacts
- Upgrade Python from 3.10 to 3.11 (matching `conda/mirage.yml`)
- Restrict `push` trigger to `mpk` branch to avoid double CI runs on PR branches

### `scripts/find_idle_gpus.sh` (new)
- Finds GPUs with 0% utilization and outputs comma-separated indices
- Supports `--num-gpus <N>` for multi-GPU selection
- Error messages go to stderr, GPU list to stdout (CI-friendly)

### `gpu-tests.yml`
- Mark as unavailable: the runs-on cloud runner (`gpu-nvidia`) is no longer functional as of April 2026
- Update push trigger to `mpk` branch

### `build-test.yml`
- Update push trigger to `mpk` branch

### `code-format.yml`
- Update push trigger to `mpk` branch

### `tests/ci-tests/run_ci_tests_qwen3.sh`
- Add `--model-path` option for loading models from local disk instead of HuggingFace

### `conda/mirage.yml`
- Add missing packages: `accelerate`, `protobuf`, `tg4perfetto`, `cuda-python`, `fastapi`, `uvicorn`

### DIFF
--- a/.github/workflows/ci-tests-qwen3.yml
+++ b/.github/workflows/ci-tests-qwen3.yml
@@ -1,6 +1,7 @@
 name: mpk-qwen-ci
 on:
   push:
+    branches: [mpk]
   pull_request:
   workflow_dispatch:
 
--- a/.github/workflows/ci-tests-qwen3.yml
+++ b/.github/workflows/ci-tests-qwen3.yml
@@ -29,7 +29,6 @@ jobs:
       - name: Setup conda environment
         run: |
           conda create -n mirage-ci-${{ github.run_id }} python=3.11 -y
-          conda activate mirage-ci-${{ github.run_id }}
 
       - name: Install mirage
         run: |
@@ -45,6 +44,7 @@ jobs:
         env:
           MIRAGE_HOME: ${{ env.MIRAGE_HOME }}
         run: |
+          conda activate mirage-ci-${{ github.run_id }}
           bash tests/ci-tests/run_ci_tests_qwen3.sh \
             --model-path /raid/catalyst/models/Qwen3-8B
 
@@ -60,6 +60,5 @@ jobs:
       - name: Clean up workspace
         if: always()
         run: |
-          conda deactivate
           conda env remove -n mirage-ci-${{ github.run_id }} -y
           rm -rf "${{ github.workspace }}/*"
--- a/.github/workflows/ci-tests-qwen3.yml
+++ b/.github/workflows/ci-tests-qwen3.yml
@@ -63,4 +63,4 @@ jobs:
         run: |
           conda deactivate
           conda env remove -n mirage-ci-${{ github.run_id }} -y
-          rm -rf "${{ github.workspace }}"
+          rm -rf "${{ github.workspace }}/*"
--- a/.github/workflows/ci-tests-qwen3.yml
+++ b/.github/workflows/ci-tests-qwen3.yml
@@ -19,6 +19,7 @@ jobs:
     env:
       CUBLAS_WORKSPACE_CONFIG: ":16:8"
       PYTHONUNBUFFERED: "1"
+      HF_HOME: "/raid/catalyst/models"
 
     steps:
       - name: Checkout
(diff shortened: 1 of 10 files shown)


conda/mirage.yml
jiazhihao, line 1: Why we have the protobuf dependency?

jiazhihao (changes requested):

conda/mirage.yml
Electronic-Waste: @jiazhihao It's included in [requirements.txt](https://github.com/mirage-project/mirage/blob/d6d38051a1408e4849a1186d4c22c8237c41c80f/requirements.txt#L10). I want to make `conda/mirage.yml` consistent with that. 

As for its usage, mirage profiler relies on it according to the doc [here](https://github.com/mirage-project/mirage/blob/d6d38051a1408e4849a1186d4c22c8237c41c80f/docs/transpiler/transpiler.md?plain=1#L271), which is also verified in the [CMakeLists.txt](https://github.com/mirage-project/mirage/blob/d6d38051a1408e4849a1186d4c22c8237c41c80f/CMakeLists.txt#L21).

jiazhihao (approved):

Electronic-Waste:
@jiazhihao Thanks for the review! I'll verify it when the fleet1 is ready:)

xinhaoc:
Since the cluster is shared by many people, let's add a timeout to the ci test and kill the process if it hangs. thanks!

.github/workflows/ci-tests-qwen3.yml
```diff
@@ -10,45 +11,43 @@ concurrency:
...
-      - name: Warm up HF cache (optional)
+      - name: Set Envs
         run: |
-          mkdir -p "$HF_HOME"
-          echo "HF cache at $HF_HOME"
-
-      - name: Set MIRAGE_HOME
-        run: echo "MIRAGE_HOME=$PWD" >> $GITHUB_ENV
+          echo "MIRAGE_HOME=$PWD" >> $GITHUB_ENV
+          CUDA_VISIBLE_DEVICES=$(bash scripts/find_idle_gpus.sh --num-gpus 1)
```
Electronic-Waste, line 43: @xinhaoc Thanks for your review! It will fail directly if it can't find an empty GPU slot. Does it sound good to you? Or shall we loop for a certain periods, like 60s?

Electronic-Waste: Or do you mean that `python demo/qwen3/demo.py --use-mirage` may hang? Can you elaborate a bit? Thanks!

xinhaoc: we can stepup a loop but it should have a max value, and what i said _time out_ is for the `demo.py `script, sometimes a PR introduces bugs and the kernel hangs and never return

xinhaoc: Yes changes in the rutime/task might make the kernel hang. for example the tma barriers might deadlock.

Electronic-Waste: Thanks for the explanation! I've added a 10-minutes timeout for this step. PTAL if you have time:)

xinhaoc: thanks! I'm not sure if the zombie process will be killed by github if it timeout.

Electronic-Waste: It will receive SIGTERM and be killed

xinhaoc: great!

xinhaoc (approved):

PR #710, opened 2026-06-08
Add online serving CI

Electronic-Waste:
## Summary
Adds a CI workflow (`mpk-online-ci`) that exercises the **online serving** path end to end: it launches the OpenAI-compatible HTTP server (`mirage.engine.launch_server`) and drives **single**, **concurrent**, and **streaming** requests against it. This complements the existing `mpk-qwen-ci`, which covers the offline token-level correctness/perf path.

## Changes
- **`.github/workflows/ci-tests-online.yml`** — new `mpk-online-ci` workflow, mirroring the Qwen3 CI conventions (`[self-hosted, gpu, b200]` runner, per-run conda env, idle-GPU selection via `find_idle_gpus.sh`, `HF_HOME`/`CUBLAS_WORKSPACE_CONFIG`, artifact upload on failure, workspace cleanup). Job timeout 60 min; serving step 40 min.
- **`tests/ci-tests/run_ci_tests_online.sh`** — launches the server in the background and owns its lifecycle: polls `/docs` until the persistent kernel finishes compiling (with a `kill -0` PID fast-fail), runs `demo_online.py` in all three modes, and tears the server down via an `EXIT` trap (TERM → grace → KILL, dumps the server-log tail on failure).
- **`demo/qwen3/demo_online.py`** — reused as the CI client. Now exits non-zero on empty/errored responses and surfaces a streamed `{"error": ...}` frame as a `RuntimeError`. Interactive behavior is unchanged.
- **`python/mirage/engine/launch_server.py`** — the streaming SSE error frame was hand-built with an f-string, producing invalid JSON whenever the error text contained a quote/backslash/newline (clients then died with a confusing `JSONDecodeError`). Now emitted via `json.dumps`.

## Testing
Validated locally without a GPU against a stdlib mock that replicates `launch_server`'s exact response/SSE shapes:
- `bash -n` / `py_compile` / YAML parse pass.
- `wait_for_ready`: ready → returns immediately; unreachable → times out; dead PID → fast-fails.
- `demo_online.py` single / concurrent×4 / streaming all pass with correct exit codes.
- Error-frame fix verified through the demo's real `_read_sse`: new format surfaces the actual server message; the old f-string format throws `JSONDecodeError`.

The full GPU run happens on the self-hosted b200 runner via this workflow.

### DIFF
--- a/.github/workflows/ci-tests-online.yml
+++ b/.github/workflows/ci-tests-online.yml
@@ -1,0 +1,68 @@
+name: mpk-online-ci
+on:
+  push:
+    branches: [mpk]
+  pull_request:
+  workflow_dispatch:
+
+concurrency:
+  group: mpk-online-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  online-serving:
+    runs-on: [self-hosted, gpu, b200]
+    timeout-minutes: 60
+    defaults:
+      run:
+        shell: bash -l {0}
+
+    env:
+      CUBLAS_WORKSPACE_CONFIG: ":16:8"
+      PYTHONUNBUFFERED: "1"
+      HF_HOME: "/raid/catalyst/models"
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          submodules: recursive
+
+      - name: Setup conda environment
+        run: |
+          conda create -n mirage-ci-${{ github.run_id }} python=3.11 -y
+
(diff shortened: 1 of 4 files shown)


Electronic-Waste:
This PR adds initial online serving CI for ensure the correctness of online serving feature. After this PR, #756 from @AravTewari can be integrated to the online serving CI instead of offline CI.

/cc @jiazhihao

jiazhihao (approved):

PR #728, opened 2026-06-15
Support DFlash for Kimi-K2.6

Risc-lt:
## Summary

This PR implements the **Kimi-K2.6 DFlash draft** in the MPK (Blackwell / sm100). Unlike Eagle3's K-step autoregressive draft, DFlash runs a **single non-causal forward per decode step**: a block of `[t0, MASK×s]` query
tokens attends non-causally to an accumulated context KV cache (plus its own block) and samples all `s` draft tokens at once.

The draft maps onto the following MPK layers per decode step (✅ reused existing kernel · 🟡 reused + new orchestration · 🔴 new kernel):

The draft maps onto the following MPK layers (✅ reused existing kernel · 🟡 reused +
new orchestration · 🔴 new kernel · ⬜ not in this PR / target-side follow-on):

| Layer | Op | What it does | Status |
|---|---|---|---|
| **L1** capture | `copy_layer` | capture the K target layers' hidden → `target_hidden [S, K·H_t]` | ⬜ target-side |
| **L7** verify | `target_verify` | verify the draft chain vs target, commit the accepted prefix | ⬜ target-side |
| **L2** fc + hidden_norm | `linear` + `rmsnorm` | project the committed `target_hidden [S, K·H_t]` → `ctx [S, H_d]` | ✅ reuse |
| **L3** input-prep | build `[t0, MASK×s]` ids + query/context positions + slot_mapping | 🟡 eager (driver-side) for now |
| **L4** materialize-KV | `linear` + `dflash_norm_rope` + `dflash_kv_store` | per draft layer: k/v proj → k_norm + YaRN-RoPE → write context K/V into the draft's paged KV cache | 🔴 new (kv_store) |
| **L5** draft forward | `rmsnorm`/`linear`/`dflash_norm_rope`/`dflash_attention`/`silu_mul` | 6 decoder layers; non-causal block attention reading the paged cache + block K/V; sliding window 2048 on L0–4, full on L5 | 🔴 new (attention) |
| **L6** sample | `narrow` + `linear` + `argmax` | gather the `s` MASK slots → target lm_head → argmax → draft tokens | ✅ reuse |

The per-step decode loop runs`target fwd → capture (L1) → verify (L7) → fc (L2) → materialize (L4) → draft (L5) → sample (L6) → next target fwd`. This PR implements the draft side (L2–L6, ⬜ L3 eager) and validates it against*captured* `target_hidden`; L1/L7 are follow-ons (see below).

## What's in the PR

**Three new sm100 kernels** (`include/mirage/persistent_kernel/tasks/blackwell/`, each wired through the full task lifecycle — enum, `task_header`, `task_register`, `graph.cc` dispatch, `runtime.cc` name map, and a `persistent_kernel.py` layer):
- **`dflash_norm_rope_sm100`** — per-head RMSNorm (eps 1e-5) + NeoX YaRN RoPE for q/k.
- **`dflash_attention_sm100`** — non-causal block attention, split context/block KV, runtime `ctx_len`, optional sliding window.
- **`dflash_kv_store_sm100`** — standalone paged KV-cache write at arbitrary slots (overwrite semantics); MPK's first KV write that isn't an attention side-effect.

**`DFlashBuilder`** (`python/mirage/mpk/models/dflash/builder.py`) — a reusable builder mirroring `Eagle3Builder` that assembles L2→L6 and owns per-layer paged KV caches accumulating committed context across decode steps. A future K2.6 target demo constructs it with the draft weights + the target's shared embed/lm_head and calls `build_step(target_hidden, query_embed, ctx_len, slot_start, num_new)`.

> YaRN detail: vLLM **and** sglang apply `mscale = 0.1·ln(64)+1 = 1.4159` for this checkpoint (their plain-`yarn` path ignores the config `mscale_all_dim` ratio that HF's `dflash.py` honors → 1.0). The draft is built at **1.4159** for parity with the deployed engines.

## Test

Validated against vLLM's DFlash output on real captured Kimi-K2.6 `target_hidden`:
- **Single decode step, full L2→L6 in one megakernel** → draft tokens **7/7 exact** vs vLLM.
- **Cross-iter to EOS, full paged-KV accumulation** (GSM8K, MPK draft fed vLLM's per-iter target capture): prompt1 **114/119 (95.8%)**, prompt2 **237/245 (96.7%)**; every miss is a bf16 near-tie (top-2 logits within ~0.05–0.5 of std ~2.1, where torch-eager and vLLM-FlashAttn also disagree).
- Per-kernel/layer: `dflash_kv_store` exact; materialize K/V relmax ~0.002; attn-from-cache with sliding window active (ctx 2104) relmax 0.0027.

The validation harness (test scripts, captured data, vLLM driver) is large/regenerable and kept local (gitignored), not in this PR.

## Follow-ons

- **Target-side loop closure — aux-hidden capture (L1) + verify (L7)**: these bracket the draft in the decode loop. Verify accepts a prefix of the draft chain and commits it; the committed tokens' captured hidden are what L2/L4 then consume.
- **True paged cache**: caches are currently a contiguous buffer with a paged-shaped tensor (slot = absolute position; no `block_table` / page pool) — correct for single-request to-EOS. Real paging (non-contiguous pages, multi-request sharing) needs a `block_table`-indirected read path; the write kernel is already paging-ready (arbitrary slots).
- **L3 input-prep** as a fused kernel (currently eager/driver-side).
- **TP=8** sharding / allreduce.

## Dataflow
image


.gitignore
xinhaoc, line 1: do we need this in gitignore?

xinhaoc (approved):

.gitignore
Risc-lt: These are just for local debugging. I can remove these if you want.

xinhaoc: yes let's remove it, and the pr is good to be merged, thanks

Risc-lt: Done. Thx for review!

PR #743, opened 2026-07-20
Deepseek v3 support

bill810975:
# Code Change
- increased num_workers for B200
- updated the num_worker schedule for qwen3 demo, so the performace and correctness consist.
- updated the kernels for Deepseek v3 tp8 ep2 design. 
- built the demo for deepseek v3.

# Performance
For Deepseek v3 tp8 ep2, `mbt=1`, batch size=1, input length = 1, tpot = 9.7ms/token.


bill810975:
This repo is ready to be reviewed. not sure if we need to keep the files in test/

PR #754, opened 2026-08-26
fix(serving): make online request lifecycle race-free

AravTewari:
## Summary

- Keep completed token rows leased until the CPU finishes copying each response.
- Use system-scope acquire/release operations for CPU/GPU handshake fields.
- Publish request-ring entries without cursor holes or completion-slot overwrite.
- Register streams before request publication, release timed-out requests, and wait for both persistent GPU streams during shutdown.

CI regression coverage is split into #756.

## Root cause

The GPU returned a completed row to the free pool before the CPU copied the response. A later request could overwrite that row while the first caller was reading it. This caused stale or cross-request chat output. Reused rows also retained stale progress state.

The CPU producer could also advance its request cursor before it published the ready flag. A concurrent flush could then publish behind an empty slot, and the GPU would stop at that permanent ring hole. Streaming registration also happened after request publication, so a fast completion could be missed.

## Validation

The six runtime-file blobs are identical to the combined revision tested on a Catalyst B200. With the functional harness from #756 applied:

- Sequential requests passed 12/12.
- Concurrent requests passed 12/12.
- HTTP status, exact response markers, cross-request isolation, streaming completion, metrics, and clean shutdown passed.
- Python compilation and `git diff --check` passed.

## Scope

- One runtime commit.
- Six runtime files.
- 365 additions and 121 deletions.
- No CI workflow or test-file changes.

### DIFF
--- a/python/mirage/mpk/persistent_kernel.py
+++ b/python/mirage/mpk/persistent_kernel.py
@@ -3170,12 +3170,10 @@ class PersistentKernel:
         self.init_func = getattr(mod, "init_func")
         self.launch_func = getattr(mod, "launch_func")
         self.init_request_func = getattr(mod, "init_request_func")
-        if self.mode == "online_pinned":
-            self.wait_func = getattr(mod, "wait_func")
+        self.wait_func = getattr(mod, "wait_func")
         self.finalize_func = getattr(mod, "finalize_func")
-        if self.mode == "online_pinned":
-            self.load_i32_acquire = getattr(mod, "load_i32_acquire")
-            self.store_i32_release = getattr(mod, "store_i32_release")
+        self.load_i32_acquire = getattr(mod, "load_i32_acquire")
+        self.store_i32_release = getattr(mod, "store_i32_release")
         
         # Prepare meta tensors
         meta_tensors = list()
--- a/python/mirage/mpk/persistent_kernel.py
+++ b/python/mirage/mpk/persistent_kernel.py
@@ -20,6 +20,7 @@ from typing import Optional
 
 HARD_CODE = """
 #include <Python.h>
+#include <cstdint>
 #include <cuda_runtime.h>
 #include <string>
 #include <vector>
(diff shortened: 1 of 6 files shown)


python/mirage/mpk/persistent_kernel.py
```diff
@@ -3113,7 +3163,12 @@ def load_mpk_kernel(
         self.init_func = getattr(mod, "init_func")
         self.launch_func = getattr(mod, "launch_func")
         self.init_request_func = getattr(mod, "init_request_func")
+        if self.mode == "online_pinned":
+            self.wait_func = getattr(mod, "wait_func")
         self.finalize_func = getattr(mod, "finalize_func")
+        if self.mode == "online_pinned":
+            self.load_i32_acquire = getattr(mod, "load_i32_acquire")
+            self.store_i32_release = getattr(mod, "store_i32_release")
```
Electronic-Waste: It's okay to make these funcs global

Electronic-Waste (commented):
Overall LGTM! Just some nits @AravTewari

Electronic-Waste:
Btw, I think it's more suitable to integrate the race CI check into the online serving CI instead of offline CI. What do you think about it? @AravTewari @jiazhihao 

I'll request review for the online serving first so that you can merge your race CI check into online serving CI.

jiazhihao:
yes, I like the idea of integrating into the online serving CI.

Electronic-Waste:
@AravTewari Great! I think the PR is ready to be merged now. @jiazhihao Can you please approve and merge it?

Btw, the CI failure is not introduced by this PR. According to the CI history, it first occurred when https://github.com/mirage-project/mirage/pull/755 is merged into the main branch.

jiazhihao (approved):

Electronic-Waste:
For documentation purpose, I list the reproduction script for race bug:

```py
#!/usr/bin/env python3
"""Reproduce the online-serving row-reuse race.

Start Mirage first:
  python -m mirage.engine.launch_server --model Qwen/Qwen3-8B \
    --max-num-batched-requests 2 --max-num-batched-tokens 4 \
    --max-seq-length 512 --request-timeout 120

Then run: python repro_online_serving_race.py [server_url]
"""

import json
import sys
import threading
import urllib.request
from concurrent.futures import ThreadPoolExecutor


URL = sys.argv[1] if len(sys.argv) > 1 else "http://127.0.0.1:8000"
BURST = 10
ROUNDS = 3


def chat(marker, gate):
    body = json.dumps({
        "messages": [{
            "role": "user",
            "content": f"Reply with exactly {marker} and nothing else.",
        }],
        "stream": False,
    }).encode()
    request = urllib.request.Request(
        f"{URL}/v1/chat/completions",
        data=body,
        headers={"Content-Type": "application/json"},
    )
    gate.wait()
    with urllib.request.urlopen(request, timeout=120) as response:
        if response.status != 200:
            raise AssertionError(f"HTTP {response.status}")
        text = json.load(response)["choices"][0]["message"]["content"]
    answer = text.rsplit("</think>", 1)[-1].strip()
    foreign_marker = text.count("MIRAGE_RACE_") != text.count(marker)
    if answer != marker or foreign_marker:
        raise AssertionError(f"{marker} received {text[-200:]!r}")


def main():
    try:
        for round_no in range(ROUNDS):
            gate = threading.Barrier(BURST)
            with ThreadPoolExecutor(max_workers=BURST) as pool:
                futures = [
                    pool.submit(chat, f"MIRAGE_RACE_{round_no:02d}_{i:02d}", gate)
                    for i in range(BURST)
                ]
                for future in futures:
                    future.result()
            print(f"round {round_no + 1}/{ROUNDS}: ok")
    except Exception as error:
        print(f"FAIL: {type(error).__name__}: {error}")
        return 1

    print("No failure reproduced. Re-run to sample another thread schedule.")
    return 0


if __name__ == "__main__":
    raise SystemExit(main())
```

Relevant discussion link: https://miragesystem.slack.com/archives/C0936E3V09W/p1787712296513899

Issue #752, opened 2026-08-14
[Bug]: Partial-argmax chunk straddling the padded-vocab boundary loses real candidates (wrong greedy token)


phact:
Companion to #751 — the second-order effect that remains after the argmax reduce is vocab-bounded. Found on commit 5c28cc68 (branch `mpk`); which decode positions break varies by GPU model because chunk widths derive from the worker count.

## Summary

`argmax_partial` tasks each scan one contiguous chunk of the padded logits and report a single (max, index) candidate. The partial kernel is position-blind. The chunk that straddles the vocab boundary contains both real and padding rows; when every real logit in that chunk is negative, its candidate is a padding row (logit 0). A vocab-bounded reduce (#751) correctly rejects that candidate; but then the chunk's real rows are unrepresented entirely, and if the global argmax lived there, the reduce returns the best of the *other* chunks: a valid but wrong token.

Concrete case: with `num_workers = 144` (170-SM GPU), chunk width = 153,600/144 ≈ 1066; the straddle chunk covers rows 151,372–152,437, which contains `<think>` (151,667). At the first decode position (all real logits negative), `<think>` was the true global argmax but was lost to the in-chunk padding candidate; the model emitted token 80711 instead.

## Proposed fix (implemented locally, demo-side)

Choose the partial-chunk width so the vocab boundary falls between chunks: `width = gcd(padded_vocab, vocab_size)` for 153,600/151,936 that's 128, giving 1,200 chunks that are each purely-real or purely-padding (the latter cleanly rejected by the bounded reduce). Three-line change in `demo/qwen3/demo.py` (partial grid dim + the two partial-result tensor widths); end-to-end per-token latency impact was within noise.

A general in-runtime alternative: give `argmax_partial` a per-task valid-row bound via task metadata so arbitrary widths become safe.

## Verification

With #751 + this: first token = `<think>` matching HF transformers greedy; 47-token exact prefix agreement on the demo prompt.

PR #758, opened 2026-08-31
fix(argmax): bound the reduction to the real vocab size

### LINKED ISSUE #751: [Bug]: Greedy argmax can emit invalid token ids: zero-padded lm_head rows win when all real logits are negative
Found on commit 5c28cc68 (branch `mpk`) while bringing MPK up on an RTX 5090, but the bug is architecture-independent.

## Summary

`demo/qwen3/demo.py` pads the lm_head weight from `vocab_size` (151,936 for Qwen3) to 153,600 rows with zeros. The padded rows therefore always produce logits of exactly 0. The argmax pipeline (`argmax_partial` --> `argmax_reduce`) scans the full padded width with no knowledge of the true vocab bound, so whenever every real logit is negative, a padding row wins and the emitted token id is ≥ vocab_size; invalid, and the decode trajectory is corrupted from there.

This is not a rare corner: the first decode position of a Qwen3 thinking-mode response is exactly such a step (HF transformers greedily picks `<think>`, id 151667, whose raw logit is negative). Any step whose top-1 logit is negative triggers it.

## Reproduction

```bash
python demo/qwen3/demo.py --use-mirage --model Qwen/Qwen3-0.6B --max-new-tokens 8 --save-tokens auto
python -c "import json; ids=json.load(open('outputs/qwen3/mpk_output.json'))['token_ids']; print(ids[0], 'INVALID' if ids[0]>=151936 else 'ok')"
```

Observed first generated id: 152243 (> 151,936). HF transformers greedy on the same prompt: 151667 (`<think>`).

## Root cause

Zero weight padding ⇒ padded logits ≡ 0; `0 > (all-negative real logits)` in the reduce. Weight-side fixes can't work because a padding row's logit is `w·h`, sign depends on the hidden state.

## Proposed fix (implemented locally, happy to PR)

Bound the reduce by the true vocab size via a defaulted template parameter (no ABI break, legacy instantiations unchanged):

- `tasks/ampere/argmax.cuh`: `template <..., long long VOCAB_SIZE = 0x7fffffffLL>`; skip candidates with `global_idx >= VOCAB_SIZE` in the reduce scan.
- `task_register.cc` `register_argmax_reduce_task`: accept an optional second param, emit the extra template arg when present.
- `persistent_kernel.py` `argmax_reduce_layer(..., vocab_size=None)`.
- demo passes `model.config.vocab_size`.

One second-order effect remains after this fix (the partial-argmax chunk that straddles the vocab boundary can still lose its real candidates to an in-chunk padding tie). Filing separately with its own fix.

With both fixes we get token-for-token greedy agreement with HF transformers for 47 tokens on the demo prompt, diverging only at a measured 0.25-logit near-tie.

Electronic-Waste:
## Summary

- Revert #755, which makes `mpk-qwen-ci` fail deterministically.
- Fix the underlying bug in the argmax task instead: reduce only over positions below the real vocab size.
- Add a test-mode regression test for the padded-lm_head case.

## Why #755 cannot work

The Qwen3 demos pad `lm_head` from `vocab_size` to 153600 rows so the partial-task grid divides evenly. `argmax_partial` reduced over the whole padded width, so a padding row could win and emit a token id `>= vocab_size` (#751 #752).

#755 tried to fix this by filling the padding **weight rows** with `-1e4`. But a padding logit is `dot(x, w_pad)`, linear in the hidden state, so a constant fill flips sign with the hidden state:

```
padding logit = dot(x, [-1e4, ..., -1e4]) = -1e4 * sum(x)
```

Whenever `sum(x) < 0` — roughly half of all decode steps — that is a huge **positive** value and argmax always picks a padding row. This is far more frequent than the all-negative-logits case #755 set out to fix. No constant weight row can give an always-negative logit, so the fix does not belong in the weight padding.

## The fix

Bound the reduction itself. `argmax_partial_kernel` / `argmax_partial_sm100_kernel` take a `VOCAB_SIZE` template argument and each task its own `chunk_start`, then reduce over `min(CHUNK_SIZE, VOCAB_SIZE - chunk_start)`. Positions at or after `vocab_size` never enter the reduction, so no padding value can win regardless of what the pad rows hold.

- `chunk_start` comes from `task_desc->task_metadata.task_offset * num_elements`, with `task_offset = bid.x` (the vocab-partitioning grid dimension) set in `runtime.cc`.
- `argmax_partial_layer(..., vocab_size=...)` passes the real vocab size as an optional `params[1]`.
- Both the template argument and the chunk offset default to the previous behaviour, so callers that do not pass `vocab_size` (llama3, deepseek_v3, glm4_moe, gpt_oss, dflash, inkling, eagle3, and the test wrapper) are unaffected.

## Validation

Run locally on a B200 with Qwen3-8B.

Differential test of #755, before vs after, same CI steps:

| | before #755 (`aa3ac171`) | at #755 (`176042f5`) |
|---|---|---|
| Torch vs MPK token comparison | pass | **fail** at position 0: `torch=151667`, `mpk=151968`; 63 of 100 MPK tokens are padding-row ids |

This matches the upstream CI history: green at `b5d01ab0`, failing from `33f35d12` onward, still failing at `cd91162b`.

With this PR, both `mpk-qwen-ci` steps pass:

- `tests/ci-tests/run_ci_tests_qwen3.sh`: `1 passed`, torch and MPK identical over the first 50 tokens, 4.03 ms/token, 7.23x over torch.
- Batch perf `r=1/2/4/8`: 225.5 / 450.1 / 896.2 / 1760.3 tokens/s, matching the pre-#755 baseline (227 / 453 / 821 / 1770) within noise, so the bound costs nothing.

New test `tests/runtime_python/test_mode/test_argmax_vocab_bound_testmode.py` builds logits where every real logit is negative and the pad region holds 0.0, covering an in-chunk winner, a chunk that straddles `vocab_size`, the last real position (151935), and a mid-vocab winner. It returns padding-row ids without the bound and the correct ids with it, and asserts each case is actually adversarial so it cannot silently degrade into a test that always passes.

## Note

`demo_sampling.py`'s `sampling_sm100_layer` is a separate token-selection path that still reduces over the padded width, so the same padding issue applies there. Left for a follow-up to keep this PR to one topic.

Fixes #751 #752

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/demo/qwen3/demo.py
+++ b/demo/qwen3/demo.py
@@ -267,7 +267,7 @@ if __name__ == "__main__":
             (
                 model.lm_head.weight,
                 torch.full(
-                    (153600 - model.config.vocab_size, hidden_size), -1e4, device="cuda"
+                    (153600 - model.config.vocab_size, hidden_size), 0, device="cuda"
                 ),
             ),
             0,
--- a/demo/qwen3/demo.py
+++ b/demo/qwen3/demo.py
@@ -754,6 +754,7 @@ if __name__ == "__main__":
             output=(argmax_part_value, argmax_part_index),
             grid_dim=argmax_partial_grid_dim,
             block_dim=(128, 1, 1),
+            vocab_size=model.config.vocab_size,
         )
         mpk.argmax_reduce_layer(
             input=(argmax_part_value, argmax_part_index),
--- a/demo/qwen3/demo_30B_A3B.py
+++ b/demo/qwen3/demo_30B_A3B.py
@@ -222,7 +222,7 @@ if __name__ == "__main__":
             (
                 model.lm_head.weight,
                 torch.full(
-                    (153600 - model.config.vocab_size, hidden_size), -1e4, device="cuda"
+                    (153600 - model.config.vocab_size, hidden_size), 0, device="cuda"
                 ),
             ),
             0,
--- a/demo/qwen3/demo_30B_A3B.py
+++ b/demo/qwen3/demo_30B_A3B.py
@@ -700,6 +700,7 @@ if __name__ == "__main__":
             output=(argmax_part_value, argmax_part_index),
             grid_dim=argmax_partial_grid_dim,
             block_dim=(256, 1, 1),
+            vocab_size=model.config.vocab_size,
         )
         mpk.argmax_reduce_layer(
             input=(argmax_part_value, argmax_part_index),
--- a/demo/qwen3/demo_30B_A3B_eagle3.py
+++ b/demo/qwen3/demo_30B_A3B_eagle3.py
@@ -236,7 +236,7 @@ if __name__ == "__main__":
             (
                 model.lm_head.weight,
                 torch.full(
-                    (153600 - model.config.vocab_size, hidden_size), -1e4, device="cuda"
+                    (153600 - model.config.vocab_size, hidden_size), 0, device="cuda"
                 ),
             ),
             0,
--- a/demo/qwen3/demo_30B_A3B_eagle3.py
+++ b/demo/qwen3/demo_30B_A3B_eagle3.py
@@ -769,6 +769,7 @@ if __name__ == "__main__":
             output=(argmax_part_value, argmax_part_index),
             grid_dim=argmax_partial_grid_dim,
             block_dim=(256, 1, 1),
+            vocab_size=model.config.vocab_size,
         )
         mpk.argmax_reduce_layer(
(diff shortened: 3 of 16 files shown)


Electronic-Waste:
This PR fixes the Qwen3 offline CI failure introduced by #755.  

/cc @jiazhihao

jiazhihao (approved):

jiazhihao:
The fix looks great to me.

PR #762, opened 2026-09-04
Serialize worker/scheduler grid launch to fix >128-worker deadlock

### LINKED ISSUE #761: [Bug] Worker counts >128 deadlock at the first request (worker/scheduler grid placement race)
MAX_NUM_WORKERS was raised to 160 for B200's 148 SMs (runtime_header.h), but any `num_workers > 128` hangs the megakernel at the first request: in split worker/scheduler mode both persistent grids are released by the same `prepare_done_event` on two concurrent streams, so their SM placement order is a per-launch race. Worker blocks need a full SM each (213KB smem); the 32-thread scheduler blocks must pack into the leftover SMs. At 128/80 there are 20 spare SMs and any order works; at 144/16 or 136/48 a wrong order strands worker blocks, and every resident worker spins forever in its dependent-event wait.

Observed on a 148-SM SM100 GPU (CUDA 13): server healthy, first request never completes; schedulers idle after dispatching one task wave, queues empty.

Fix in the linked PR: gate the scheduler-grid launch on a device counter reaching `num_workers` (each worker block increments once at entry), with a 60s timeout diagnostic instead of a silent hang. Zero steady-state cost.

Separate note: reaching >128 workers on Qwen3 also requires decoupling the sampling-partition count from `num_workers` (padded vocab 153600 % 144 != 0 trips the divisibility assert); models whose padded vocab divides (e.g. Llama-3, 129024) reach the deadlock directly.

jyizheng:
Fixes #761.

In split worker/scheduler mode both persistent grids become runnable on the same `prepare_done_event` on two concurrent streams, so SM placement order is a per-launch race; with fewer than ~20 spare SMs (any `num_workers > 128` on a 148-SM part) a wrong order strands worker blocks and the megakernel deadlocks in the dependent-event wait.

Fix: worker blocks increment a device counter at entry; the scheduler grid launch waits for it to reach `num_workers` (60s timeout with a loud diagnostic instead of a silent hang). Purely additive, zero steady-state cost, existing configs unchanged.

Validated on a 148-SM SM100 GPU with this exact patch: 144/16 serves requests and completes a 20-step group-decode run; 128/80 unchanged (no regression).

### DIFF
--- a/include/mirage/persistent_kernel/persistent_kernel.cuh
+++ b/include/mirage/persistent_kernel/persistent_kernel.cuh
@@ -1400,6 +1400,11 @@ __global__ __launch_bounds__(WORKER_NUM_THREADS,
 
 __global__ __launch_bounds__(WORKER_NUM_THREADS,
                              1) void worker_kernel(RuntimeConfig config) {
+  // Report residency so the host can gate the scheduler-grid launch on all
+  // worker blocks being placed (see launch_persistent_kernel).
+  if (threadIdx.x == 0) {
+    atomicAdd(config.worker_block_arrival_count, 1ULL);
+  }
   worker_checker(config);
   execute_worker(config);
 }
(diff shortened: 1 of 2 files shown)


jyizheng:
Could any body review this PR?

Electronic-Waste:
@jyizheng Sorry for the late reply. Our maintainers' bandwidths are limited this semester so we missed your PR.

I think the issue you raise is great! We also encounter this scenarios from time to time. As for the fix, I think the global counter can introduce more synchronization and thus harm the performance. So, we decide to adopt the fix in #775 which only modify the gridDim/blockDim in the launch phase of scheduler kernel.

Anyway, thanks for your great contribution. And welcome to the mirage community!

cc @jiazhihao @xinhaoc

jyizheng:
#775 landed and fixes the root cause with a lighter approach — closing this one. Thanks @Electronic-Waste for the pointer, and glad the >128-worker deadlock is resolved.

PR #768, opened 2026-09-14
Fix Hopper multitoken_paged_attention mbt>1 stale-tail NaN

### LINKED ISSUE #740: [Feature Request] - supporting larger --max-num-batched-tokens, especially on the MoE path
### Feature Description

Hi, thanks for the great work on MPK!

While studying the batching behavior I noticed the following note in run_tgx.sh (lines 58–60):
NOTE: do NOT pass --max-num-batched-tokens. Each demo's default is what
its kernel was tuned for (qwen3 hopper demos: 8; 30B-A3B MoE: 1). Forcing
8 on the MoE path silently produces zeros/eos sentinel in the output.

This lines up with what I observe in benchmarks: MPK's advantage degrades once the number of requests exceeds --max-num-batched-tokens.

Dense models (e.g. Qwen3 Hopper demos, default max-num-batched-tokens = 8): MPK handles up to 8 requests concurrently, so its advantage starts to shrink around batch size = 16.
MoE models (e.g. 30B-A3B, default max-num-batched-tokens = 1): only 1 request can be processed at a time, so the advantage decreases steadily for any batch size > 1.

My questions:

（1）Why not tune/adapt the kernels for a larger --max-num-batched-tokens? Is this a deliberate design choice or a work-in-progress?

（2）Is there a fundamental limitation behind this — e.g. register file capacity per SM, shared-memory budget per SM — that makes larger batches hard to support?

（3）Specifically for the MoE path: what causes the "silently produces zeros/eos sentinel" failure when max-num-batched-tokens is forced above 1? Is it a kernel correctness issue that could be fixed, or an inherent constraint?

Any pointers to the relevant code or design rationale would be very helpful. Thanks!

jyizheng:
The Hopper KV prefetch loop bounded its copies by the valid row count (curr_iter_len / next_iter_len), leaving the tail rows of the KV tile with stale shared memory. `P @ V` reads the whole tile after masking `P`, and a masked probability of zero does not suppress a stale NaN in V (0 * NaN = NaN), so the tail NaN propagates into every output row once the context length is not a multiple of KV_TILE_SIZE, at `--max-num-batched-tokens > 1`. On Qwen3-30B-A3B this surfaces as out-of-range argmax tokens (demo_30B_A3B_hopper.py --max-num-batched-tokens 8).

Load the full tile and zero the tail rows; add the producer barrier before publishing the tile and restore the consumer barrier before the output gather. Hopper counterpart of the sm100 prefetch-range fix (#760).

Test: tests/runtime_python/hopper/test_attention_tail_hopper.py over (capacity, active, context) grids — all NaN before, finite (max_error ~1e-3) after; the 30B demo at mbt=8 goes from OverflowError to coherent output.

Fixes #740.


jyizheng:
@Electronic-Waste While you're catching up on the queue — this one fixes a Hopper mbt>1 stale-tail NaN in multitoken_paged_attention, CI is green. Mind taking a look?

Electronic-Waste:
@jyizheng Yes, sure. I'll take a look in these days. Sorry for the late reply:)

Electronic-Waste (approved):
@jyizheng Thanks for the great contribution! This PR LGTM!

Additional context: overhead of the extra barriers and full-tile loads at mbt=1: 5.404 → 5.422 ms, which is acceptable.

PR #774, opened 2026-09-17
test: allow test_mode to run against a pre-seeded KV cache

JerryLiang49:
Today every test_mode attention test can only run a plain prefill, where the whole sequence is new. The runtime builds the batch itself before the first task graph runs, and it overwrites any batch a test tries to supply. So the kernel always sees a sequence with no history behind it.

That leaves two things untested:
- Decode, and chunked prefill, reading keys and values that an earlier iteration wrote into the cache.
- The sliding window skipping whole leading KV tiles. The kernel only skips tiles when the sequence is longer than the new tokens, which never happens today

### Change
A new opt-in flag, `test_inject_batch`. When it is on, the runtime keeps the batch the test provides and runs it for exactly one task graph instead of building its own. It is off by default, so every existing test behaves the same as before.


Electronic-Waste (commented):
@JerryLiang49 Thanks for your great contribution! The idea is great.

I'm not sure if we need to introduce more flags in the main function. As you can see, there're plenty of flags now. But I would like to hear from your opinions @JerryLiang49 @jiazhihao @xinhaoc

JerryLiang49:
@Electronic-Waste Good point. I see that the list of compile flags is already long.

I could remove the compile flag itself. Instead of `MPK_TEST_INJECT_BATCH`, the setting could be a field in `RuntimeConfig,` with `init_kernel` and the scheduler checking it at runtime. 

The other option is to treat it as another serving mode. `prepare_next_batch()` already has its own path for `MODE_OFFLINE`, `MODE_ONLINE`, `MODE_ONLINE_PINNED` and `MODE_ONLINE_NOTOKEN`.

Electronic-Waste:
So, if I understand correctly, what you mean "test_mode" is initializing KV cache first before a test case starts. Right?

In this case, I do not see the necessity to add more flags/modes to the main branch. Although we do not test decode and chunked prefill in unit tests, we've covered the scenarios you mentioned in the offline e2e CI.

If you do like to add pre-seeded KV cache for testing, I would recommend to referencing vllm-style test cases, which constructs KV cache directly in test files with `torch.randn`: https://github.com/vllm-project/vllm/blob/5302d1fe40e6969b854e0ab288d34825653767c6/tests/kernels/attention/test_flash_attn.py#L255-L257

And thanks for your clarification! @JerryLiang49

JerryLiang49:
You're right I must have missed this. The flag isn't needed so I'll close this PR.

PR #775, opened 2026-09-17
fix: prevent split-launch deadlocks by packing schedulers

phamson02:
## Problem

A scheduler block can occupy an SM before a worker starts. Workers require most of an SM's shared memory and cannot reliably share an SM with a running scheduler. If scheduler blocks occupy too many SMs, some workers cannot start. Tasks assigned to those workers never complete and cause a deadlock.

This can happen because there is no guaranteed execution order between the worker and scheduler kernels, which run on separate streams. The launch also creates more blocks than SMs. The default sizing policy assumes one worker per SM and four schedulers per SM, but the launch creates one block per scheduler. Four scheduler blocks can therefore occupy four SMs instead of one and prevent some workers from starting.

On a B200 with 148 SMs, the default configuration selects 128 workers and 80 schedulers. When the scheduler blocks are placed first, they occupy 80 distinct SMs, and only 68 of the 128 workers start. The demo then hangs after "Finished Launching Persistent Kernel (Async)".

## Change

Pack exactly four scheduler warps into each scheduler block so that they run on the same SM. Use `SCHEDULERS_PER_BLOCK` for launch dimensions and scheduler ID calculations, including the kernel that runs both workers and schedulers.

For default counts, the scheduler blocks fit within the reserved SM budget even when they start first. On a B200, the 80 schedulers use 20 blocks and leave 128 SMs available for workers. This removes the dependence on worker blocks starting first.

Custom scheduler counts must be divisible by four. The separate kernels use the local scheduler count, while the single worker/scheduler kernel uses the total local and remote count. Custom configurations still need enough SMs for both workers and schedulers.

An alternative in #762 waits for every worker to report entry before launching schedulers. Packing avoids the worker-entry counter and CPU polling. This change adds no runtime state.

## Reproduction

The hang depends on which grid the hardware places first, so a plain run passes or hangs by chance. To make the scheduler grid land first every time, edit `launch_persistent_kernel` in `include/mirage/persistent_kernel/persistent_kernel.cuh`: move the `scheduler_kernel<<<...>>>` launch above the `worker_kernel<<<...>>>` launch and add a short host sleep between them.

```cpp
    scheduler_kernel<<<...>>>(global_runtime_config);   // moved up
    std::this_thread::sleep_for(std::chrono::milliseconds(5));
    worker_kernel<<<...>>>(global_runtime_config);
```

`<thread>` is already included; add `#include <chrono>` next to it. Then run the demo on one B200:

```bash
python demo/qwen3/demo.py --use-mirage
```

Before this fix, the run prints "Finished Launching Persistent Kernel (Async)" and never returns. With this fix and the same reordering, the run completes and prints the per-token latency.

Placement can be confirmed by reading `%smid` in each block at kernel entry. In the hung run, 80 scheduler blocks occupy 80 distinct SMs and 68 of 128 workers start. With the fix, 80 scheduler warps occupy 20 SMs and all 128 workers start.

## Validation

Runtime tests use a B200 with CUDA 12.8 and driver 580.95.05, the default configuration of 128 workers and 80 schedulers, and the Qwen3-8B demo.

With the scheduler grid forced to place first, the old launch puts 80 scheduler blocks on 80 distinct SMs, only 68 of 128 workers start, and the demo hangs. The packed launch puts the same 80 schedulers on 20 SMs, all 128 workers start, and the demo completes. Each case was run twice with the same result. Without the forced order, both launches place identically and complete.

Linked Issues:
- Issue #761


include/mirage/persistent_kernel/persistent_kernel.cuh
```diff
@@ -1833,11 +1833,20 @@ extern "C" void launch_persistent_kernel(cudaStream_t default_stream) {
...
 
-    scheduler_kernel<<<dim3(global_runtime_config.num_local_schedulers, 1, 1),
-                       dim3(32, 1, 1),
-                       0 /*smem*/,
-                       global_runtime_config.scheduler_stream>>>(
-        global_runtime_config);
+    // Pack scheduler warps into blocks so they leave enough SMs for workers,
+    // even when CUDA starts the schedulers before the workers.
+    int schedulers_per_block =
+        std::min(4, global_runtime_config.num_local_schedulers);
```
Electronic-Waste: Can you make `schedulers_per_block` a global variable in `runtime_header.h`? I think it will be more general. WDYT?

include/mirage/persistent_kernel/persistent_kernel.cuh
```diff
@@ -1833,11 +1833,20 @@ extern "C" void launch_persistent_kernel(cudaStream_t default_stream) {
...
+    while (global_runtime_config.num_local_schedulers % schedulers_per_block) {
+      --schedulers_per_block;
+    }
```
Electronic-Waste: It's very tricky to adjust `schedulers_per_block` here. We can just verify it with `assert` in functions like `init_persistent_kernel` instead of adjusting it there.

Electronic-Waste (commented):
@phamson02 Thanks for your great contribution! And welcome to the mirage community.

The PR overall LGTM. I left some comment for you.

Also cc @jiazhihao @xinhaoc

include/mirage/persistent_kernel/persistent_kernel.cuh
phamson02: Agree. I will do that

include/mirage/persistent_kernel/persistent_kernel.cuh
phamson02: Agree

phamson02:
Updated based on @Electronic-Waste's comments. cc @jiazhihao @xinhaoc

include/mirage/persistent_kernel/persistent_kernel.cuh
```diff
@@ -804,10 +804,10 @@ __device__ __forceinline__ void persistent_checker(RuntimeConfig config) {
   assert(gridDim.y == 1);
   assert(gridDim.z == 1);
   // Each worker SM serves a single worker
-  // Each scheduelr SM serves four schedulers
+  // Each scheduler SM serves four schedulers
   int const num_schedulers =
       config.num_local_schedulers + config.num_remote_schedulers;
-  int const num_schedulers_per_sm = std::min((int)blockDim.x / 32, 4);
+  int const num_schedulers_per_sm = SCHEDULERS_PER_BLOCK;
```
Electronic-Waste, line 810: Can you preserve these code? It will be better if we could make `num_scheduler_per_sm` adaptive.

include/mirage/persistent_kernel/persistent_kernel.cuh
```diff
@@ -1115,14 +1115,12 @@ __device__ __forceinline__ void execute_worker(RuntimeConfig config) {
...
-// need to alter as there is only one warp per block
 __device__ __forceinline__ void execute_scheduler(RuntimeConfig config,
                                                   int offset) {
   int const num_schedulers =
       config.num_local_schedulers + config.num_remote_schedulers;
-  // if we have more than 4 warps per thread block
-  // only the first 4 warps will run schedulers
-  int const num_schedulers_per_sm = std::min((int)blockDim.x / 32, 4);
+  // Only the first SCHEDULERS_PER_BLOCK warps run schedulers.
+  int const num_schedulers_per_sm = SCHEDULERS_PER_BLOCK;
```
Electronic-Waste, line 1123: Same as above

include/mirage/persistent_kernel/persistent_kernel.cuh
```diff
@@ -1632,6 +1630,12 @@ extern "C" void
   global_runtime_config.my_gpu_id = mype;
   global_runtime_config.num_graphs = 1;
   global_runtime_config.split_worker_scheduler = true;
+  // Split launch is currently forced. Keep both cases so switching to combined
+  // launch validates the total scheduler count.
+  assert((global_runtime_config.split_worker_scheduler ? num_local_schedulers
+                                                       : num_schedulers) %
+             SCHEDULERS_PER_BLOCK ==
+         0);
```
Electronic-Waste:
```suggestion
```

Electronic-Waste: It's okay to just verify split mode for now since non-split mode is deprecated for now.

Electronic-Waste (commented):
@phamson02 LGTM. Just a few comments for you

include/mirage/persistent_kernel/persistent_kernel.cuh
Electronic-Waste: And we can place the assert logic at the beginning of `init_persistent_kernel`

include/mirage/persistent_kernel/persistent_kernel.cuh
phamson02: I interpret this as more like a request to keep the kernel's code intact, since I am not sure how `num_scheduler_per_sm` can be adaptive here. Can you elaborate further here?

Electronic-Waste: Oh, you're right. I made a mistake.

In the former implementation, we derives `num_scheduler_per_sm` with `std:min` because `execute_scheduler` is trigger by two mode:

1. Non-split: 4 scheduler warps fused with worker warps in a single block
2. Split: 1 scheduler warp each block (`blockDim.x=32` -> `num_schedulers_per_sm=1`)

In the current fix against the race bug in split mode, we fuse 4 scheduler warp in a block and there's no different between the `num_scheduler_per_sm` of split and non-split mode. So we can just use a unified `SCHEDULER_PER_BLOCK`. You're right.

phamson02: Thanks for the clarification 😉

phamson02:
Updated

Electronic-Waste (approved):
@phamson02 LGTM! Thanks for your great contribution!

PR #777, opened 2026-09-17
Initialize Qwen3 MPK state buffers

### LINKED ISSUE #776: [Bug] - Qwen3 MPK produces different tokens across processes with uninitialized KV and metadata buffers
## Bug Description

The Qwen3 MPK demo allocates its KV cache and paged-KV metadata buffers with
`torch.empty()`. On NVIDIA B200, separate process executions with identical
model weights, prompt, and runtime settings generated different output tokens.

Changing the affected allocations from `torch.empty()` to `torch.zeros()`
eliminated the cross-process nondeterminism in our tests.

The behavior was originally tested at commit
`c9d83035d2e79cf85ae6a1e5dcbb507c4f6bbe61`. The same allocation pattern is
still present on the current `mpk` branch at
`1f3338f9ae084559726fb1fc898ae077ff7d171a`.

### Affected tested path

`demo/qwen3/models/modeling_qwen3.py` allocates the key and value caches with
`torch.empty()`:

```python
key_cache = torch.empty(...)
value_cache = torch.empty(...)
```

`demo/qwen3/demo.py` allocates the paged-KV metadata with `torch.empty()`:

```python
qo_indptr_buffer = torch.empty(...)
paged_kv_indptr_buffer = torch.empty(...)
paged_kv_indices_buffer = torch.empty(...)
paged_kv_last_page_len_buffer = torch.empty(...)
```

These tensors are attached to the generated persistent kernel and become raw
GPU pointers used during inference.

### Observed behavior

The test used Qwen3-0.6B with fixed weights and prompt:

- 28 layers
- one request
- maximum sequence length 64
- page size 64
- four pages
- 128 workers and 80 schedulers

Two separate process executions completed the same number of decode steps but
generated different tokens:

```text
run 1 output_tokens_sha256:
b566cec9d24976ec42756e25a806247bc19686593537650654564afb02520dd2

run 2 output_tokens_sha256:
45896661fdae674d6b90ef46fe224ebc8b61a7da947c28e249148098b9a8fcac
```

The `step_sha256` value was identical in both runs, indicating that the
execution completed the same decode length while the token values differed.

After zero-initializing the two KV caches and four metadata buffers, the
recorded cross-process runs produced identical output hashes.

This is consistent with the persistent-kernel path reading CUDA allocator state
from a buffer that has not been completely initialized before use.

### Proposed fix

Zero-initialize state that may be read before every element is explicitly
written:

```diff
- torch.empty(...)
+ torch.zeros(...)
```

The tested patch changes the two Qwen3 KV-cache allocations in
`demo/qwen3/models/modeling_qwen3.py` and the four paged-KV metadata allocations
in `demo/qwen3/demo.py`.

Similar `torch.empty()` allocation patterns are present in the Qwen3 MPK builder
and fallback metadata allocation paths. Those paths may warrant the same audit,
but the result above is specifically for the Qwen3 demo path.

## Build Environment

- NVIDIA B200
- CUDA 12.9
- Mirage branch `mpk`
- Tested commit `c9d83035d2e79cf85ae6a1e5dcbb507c4f6bbe61`

The current environment available to us uses NVIDIA H100, so the B200
experiment was not rerun against the latest commit.

## Error Message Output

There is no error message or crash. The issue manifests as silent differences
in generated tokens across separate process executions with fixed inputs.

This issue was identified, analyzed, and documented during a fine-grained,
interactive debugging session with GitHub Copilot, powered by GPT-5.6. The work
was conducted as part of a research project on GPU testing and verification
that is currently under submission.

tyler-utah:
**Description of changes:**

- zero-initialize the Qwen3 demo key and value caches
- zero-initialize the Qwen3 demo paged-KV metadata buffers
- prevent stale CUDA allocator contents from affecting persistent-kernel inference

On NVIDIA B200, separate process executions with fixed model weights, prompt,
and runtime settings generated different output tokens when these buffers were
allocated with `torch.empty()`. Changing these six allocations to
`torch.zeros()` made the recorded cross-process runs match.

This PR intentionally changes only the Qwen3 demo path used in the reported
experiment. Similar fallback and builder allocations can be audited separately
without broadening this fix beyond the validated scope.

The B200 experiment was conducted at commit
`c9d83035d2e79cf85ae6a1e5dcbb507c4f6bbe61`. The same allocation pattern remains
on the current `mpk` branch. The available system now uses NVIDIA H100, so the
B200 result was not rerun against the latest commit.

This issue and fix were identified, analyzed, and documented during a
fine-grained, interactive debugging session with GitHub Copilot, powered by
GPT-5.6. The work was conducted as part of a research project on GPU testing
and verification that is currently under submission.

**Related Issues:**

Linked Issues:
- Issue #776

Issues closed by this PR:
- Closes #776


Electronic-Waste:
@tyler-utah Thanks for your great contribution! And welcome to the mirage community:)

We've met with this mismatch for a long time but haven't got a chance to fix it. However, I think it might be a non-determinism problem rather than KV cache pollution because the llm engine first writes KV cache then read it. But in any case, "ensuring determinism in megakernel" is an interesting idea and truly worthwhile to explore.

Please tell me how you think of it!
 
Also cc @jiazhihao @xinhaoc

tyler-utah:
Hi @Electronic-Waste,

Thanks for the nice message and for looking into the issue. After looking more into the code, we found that our testing framework had an issue and you are absolutely correct.

After fixing the testing framework (and firing up GPT 6), we found three issues. Addressing them gave us reliably deterministic results in our H100 tests:

1. The [V-loading loop (lines 316–337)](https://github.com/mirage-project/mirage/blob/bc5d69612a345efe502a4dfb28faa672dd35dbeb/include/mirage/persistent_kernel/tasks/hopper/multitoken_paged_attention_hopper.cuh#L316-L337) leaves tile padding uninitialized. In a failing run, residual shared-memory contents included NaNs that propagated through the matrix multiply despite zero attention weights. Zero-initializing the padding eliminated that failure.

2. It looks like a barrier was already [present (lines 724–725)](https://github.com/mirage-project/mirage/blob/bc5d69612a345efe502a4dfb28faa672dd35dbeb/include/mirage/persistent_kernel/tasks/hopper/multitoken_paged_attention_hopper.cuh#L724-L725), but was commented out for some reason. Our tests indicate that it is necessary: re-enabling it eliminated one source of nondeterministic results.

3. The [initial K/V load (lines 338–344)](https://github.com/mirage-project/mirage/blob/bc5d69612a345efe502a4dfb28faa672dd35dbeb/include/mirage/persistent_kernel/tasks/hopper/multitoken_paged_attention_hopper.cuh#L338-L344) can signal that the first tile is ready before all producer warps finish their copies. Adding the producer barrier already [used for subsequent tiles (lines 386–387)](https://github.com/mirage-project/mirage/blob/bc5d69612a345efe502a4dfb28faa672dd35dbeb/include/mirage/persistent_kernel/tasks/hopper/multitoken_paged_attention_hopper.cuh#L386-L387) eliminated further numerical mismatches and run-to-run variation.

We also benchmarked the changes. In short-sequence Qwen3-0.6B tests on H100, the first two fixes together added about 1.4–2.1% to e2e inference latency, and the third barrier had negligible additional overhead.

Any thoughts on the above? I can make another PR if you are interested.

Electronic-Waste:
Yes, we're absolutely interested in this feature! Batch invariance is important in both training and inference. 

As for the performance loss, I think it's acceptable. Any ideas on it? @jiazhihao @xinhaoc

tyler-utah:
Thanks for the reply @Electronic-Waste!  I'll prepare a PR then; it is a small amount of changes.

Just to clarify though, I don't believe this has to do with batch invariance. In our tests, batch sizes and seeds are all kept stable. The issues I'm writing about seem to be more traditional correctness issues, e.g., uninitialized variables and race conditions. 

We're excited to share more about our work with you soon! Hopefully you can find our tools useful

tyler-utah:
I just noticed #768 already addresses these same three issues. Our padding fix was slightly different, but it should behave the same.

I'll just point to that PR instead of opening new ones.

Thanks!

